PR #52, opened 2021-07-15
Add spec for URLPattern getters.

wanderview:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/52.html" title="Last updated on Jul 15, 2021, 6:51 PM UTC (db04c07)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/52/ed998c5...db04c07.html" title="Last updated on Jul 15, 2021, 6:51 PM UTC (db04c07)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -83,7 +83,7 @@ table {
 
 <script src="https://resources.whatwg.org/file-issue.js" async></script>
 
-<h2 id="global">The {{URLPattern}} class </h2>
+<h2 id=urlpattern-class>The {{URLPattern}} class </h2>
 
 <xmp class="idl">
 typedef (USVString or URLPatternInit) URLPatternInput;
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic can you review this PR to add initial support for the URLPattern getters?  Note, I plan to add more information to the `component` object in future PRs.  Just FYI in case you were wondering why it was an object with one piece of information on it, etc.

spec.bs
```diff
@@ -136,3 +136,115 @@ dictionary URLPatternComponentResult {
   record<USVString, USVString> groups;
 };
 </xmp>
+
+Each {{URLPattern}} object has an associated <dfn for=URLPattern>protocol component</dfn>, a <a for=/>component</a>, initially null.
```
domenic, line 140: If you want you can use `[=/component=]` here and below; `<a>` should never really be necessary.

spec.bs
```diff
@@ -136,3 +136,115 @@ dictionary URLPatternComponentResult {
...
+Each {{URLPattern}} object has an associated <dfn for=URLPattern>pathname component</dfn>, a <a for=/>component</a>, initially null.
+
+Each {{URLPattern}} object has an associated <dfn for=URLPattern>search component</dfn>, a <a for=/>component</a>, initially null.
+
+Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component</dfn>, a <a for=/>component</a>, initially null.
+
+<dl class="domintro non-normative">
+  <dt><code>{{URLPattern}} . {{URLPattern/protocol}}</code>
+  <dd>
+    <p>The normalized protocol pattern string.
```
domenic, line 159: Web developers might enjoy these being cross-linked to the appropriate parts of the URL spec, e.g. `The normalized [=URL/scheme=] pattern string`. But I'm not sure.

spec.bs
```diff
@@ -136,3 +136,115 @@ dictionary URLPatternComponentResult {
...
+
+<div algorithm>
+  The <dfn attribute for="URLPattern">hash</dfn> getter steps are:
+
+  1. Return [=this=]'s [=URLPattern/hash component=]'s [=component/pattern string=].
+</div>
+
+<h2 id=components>Components</h2>
+
+<p>A {{URLPattern}} is associated with multiple <dfn id=concept-component>component</dfn> objects.
```
domenic, line 248: Manually curating `id=""` values can be kind of annoying. I'd personally suggest not doing so and just accepting the auto-assigned IDs, i.e. omit the `id=""` attribute. Otherwise you'll either end up with a lot of work to do this for every definition you introduce, or with inconsistency because you only do it sometimes.

domenic (approved):

spec.bs
wanderview: Rewrote to use `[=component=]`.  Note, I didn't seem to need that slash in there.

spec.bs
wanderview: Hmm, I tried this but its failing with "LINK ERROR: No 'dfn' refs found for 'scheme' with for='['URL']'."  Do I need to include it somehow?

wanderview: Oh, it appears to need lowercase `url/scheme`

wanderview: Hmm, I think I'm not going to do this.  The names don't quite match up (scheme-vs-protocol, search-vs-query, etc).  Also, if you then follow the links they give you definitions for how they should be defined that are not true for pattern strings.

spec.bs
wanderview: Done

spec.bs
domenic: Yeah, the latter was my main uncertainty. Makes sense.

PR #53, opened 2021-07-15
Spec matching methods except for init canonicalization.

wanderview:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/53.html" title="Last updated on Jul 16, 2021, 8:19 PM UTC (83983ed)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/53/902b1aa...83983ed.html" title="Last updated on Jul 16, 2021, 8:19 PM UTC (83983ed)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -18,6 +18,21 @@ Default Biblio Status: current
 Markup Shorthands: markdown yes
 </pre>
 
+<pre class="link-defaults">
+spec:infra; type:dfn; text:list
+spec:webidl; type:dfn; text:record
+</pre>
+
+<pre class="anchors">
+spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
+  type: abstract-op
+    text: Get; url: #sec-get-o-p
+    text: RegExpBuiltinExec; url: #sec-regexpbuiltinexec
+    text: ToString; url: #sec-tostring
+  type: interface
+    text: RegExp; url: #sec-regexp-regular-expression-objects
+</pre>
+
 <style>
 .selected-text-file-an-issue {
   position: fixed;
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic can you review this PR?  Its a bit bigger and does most of the matching algorithm.  It does not do canonicalization of init members yet, though, because I need to figure out how to integrate with URL spec for that.

For reference, the c++ code for this can be found here:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/url_pattern/url_pattern.cc;l=558;drc=4d354dd4d0fa1db2f924262a4abe67e2fda9d7b4

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+
+  1. Let |result| be the result of [=match=] given [=this=], |input|, and |baseURL|.
+  1. If |result| is null, return false.
+  1. Return true.
+</div>
+
+<div algorithm>
+  The <dfn method for="URLPattern">exec(|input|, |baseURL|)</dfn> method steps are:
+
+  1. Return the result of [=match=] given [=this=], |input|, and |baseURL|.
```
domenic, line 264: Here and above, to make it clear you're passing along the optionality, I'd say `and |baseURL| if given`

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+
+  1. Let |protocol| be the empty string.
+  1. Let |username| be the empty string.
+  1. Let |password| be the empty string.
+  1. Let |hostname| be the empty string.
+  1. Let |port| be the empty string.
+  1. Let |pathname| be the empty string.
+  1. Let |search| be the empty string.
+  1. Let |hash| be the empty string.
+  1. Let |inputs| be an empty array.
```
domenic, line 278: In spec-land we use [lists](https://infra.spec.whatwg.org/#lists) so as not to be confused with developer-modifiable JavaScript arrays. So you'd use `an empty [=list=]` and `[=list/Append=]` instead of `Push` below.

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Push |input| on to |inputs|.
+  1. If |input| is a {{URLPatternInit}} then:
+    1. If |baseURLString| is not null, throw a TypeError.
```
domenic, line 281: As you've stated things now it's "optional", not nullable. So the right phrasing is `If |baseURLString| was given`.

domenic:
```suggestion
    1. If |baseURLString| is not null, throw a {{TypeError}}.
```

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+    1. Let applyResult be the result of [=process a URLPatternInit=] given |input|, "url", |protocol|, |username|, |password|, |hostname|, |port|, |pathname|, |search|, and |hash|. If this throws an exception, catch it, and return null.
```
domenic, line 282:
```suggestion
    1. Let |applyResult| be the result of [=process a URLPatternInit=] given |input|, "`url`", |protocol|, |username|, |password|, |hostname|, |port|, |pathname|, |search|, and |hash|. If this throws an exception, catch it, and return null.
```

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+    1. Set |protocol| to |applyResult|'s {{URLPatternInit/protocol}}.
```
domenic, line 283:
```suggestion
    1. Set |protocol| to |applyResult|["{{URLPatternInit/protocol}}"].
```

(here and elsewhere... there are a ton of instances of this :))

We treat dictionaries as [ordered maps](https://infra.spec.whatwg.org/#ordered-map) so we use this syntax for accessing their members.

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+    1. Set |search| to |applyResult|'s {{URLPatternInit/search}}.
+    1. Set |hash| to |applyResult|'s {{URLPatternInit/hash}}.
+  1. Else:
+    1. Let |baseURL| be null.
+    1. If |baseURLString| is not null then:
+      1. Set |baseURL| to the result of [=URL parser|parsing=] |baseURLString|.
+      1. If |baseURL| is failure, return null.
+      1. Push |baseURL| on to |inputs|.
+    1. Let |url| be the result of [=URL parser|parsing=] |input| given |baseURL|.
+      1. If |url| is failure, return null.
```
domenic, line 298:
```suggestion
    1. If |url| is failure, return null.
```

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+    1. Set |protocol| to |url|'s {{URL/protocol}}.
```
domenic, line 299: It's not great practice to refer directly to object properties like `protocol`, since they are modifiable by JavaScript (e.g. JS messing with `URL.prototype`).

Additionally, |url| is a [URL record](https://url.spec.whatwg.org/#concept-url), not a `URL` instance.

This is tricky. I see a few options:

1. Duplicate all the getter steps logic here for the components you're interested in.
2. Work on a refactoring of the URL spec to abstract these all out into algorithms you can call, e.g. "the API hostname string" for a URL record.
3. Try to sidestep the issue by creating Web IDL objects and referencing them.

(3) is somewhat unprecedented but perhaps easiest. It would look something like:

```
1. Let |urlObject| be a [=new=] {{URL}}.
2. Set |urlObject|'s [=URL/URL=] to |url|.
3. Set |protocol| to the result of applying the getter steps for {{URL/protocol}} to |urlObject|.
```

Having written this out I do not like this very much. I would suggest (2). What that would look like for now would be adding a section at the bottom of your spec which says something like

```html
This spec depends on factoring out various getter steps for the {{URL}} object into new algorithms that operate on URL records and return [=scalar value strings=]. These are the <dfn for="url">API protocol string</a>, the <dfn for="url">API username string</a>, ...
```

i.e. you don't have to do the work of actually factoring things out at this stage; just rely on it being rather obvious.

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+    1. Set |username| to |url|'s {{URL/username}}.
+    1. Set |password| to |url|'s {{URL/password}}.
+    1. Set |hostname| to |url|'s {{URL/hostname}}.
+    1. Set |port| to |url|'s {{URL/port}}.
+    1. Set |pathname| to |url|'s {{URL/pathname}}.
+    1. Set |search| to |url|'s {{URL/search}}.
+    1. Set |hash| to |url|'s {{URL/hash}}.
+  1. Let |protocolExecResult| be the result of [=RegExpExec=] given |urlpattern|'s [=URLPattern/protocol component=]'s [=component/regular expression=] and |protocol|.
```
domenic, line 307: You want RegExpBuiltinExec. Otherwise modifying `RegExp.prototype.exec` will affect your results.

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. If |protocolExecResult| is null then return null.
```
domenic, line 308: Something that might make the spec easier to read is if you grouped all these steps at the end. `If |protocolExecResult|, |usernameExecResult|, |passwordExecResult|, ..., or |hashExecResult| is null, then return null`.

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. If |hostnameExecResult| is null then return null.
+  1. Let |portExecResult| be the result of [=RegExpExec=] given |urlpattern|'s [=URLPattern/port component=]'s [=component/regular expression=] and |port|.
+  1. If |portExecResult| is null then return null.
+  1. Let |pathnameExecResult| be the result of [=RegExpExec=] given |urlpattern|'s [=URLPattern/pathname component=]'s [=component/regular expression=] and |pathname|.
+  1. If |pathnameExecResult| is null then return null.
+  1. Let |searchExecResult| be the result of [=RegExpExec=] given |urlpattern|'s [=URLPattern/search component=]'s [=component/regular expression=] and |search|.
+  1. If |searchExecResult| is null then return null.
+  1. Let |hashExecResult| be the result of [=RegExpExec=] given |urlpattern|'s [=URLPattern/hash component=]'s [=component/regular expression=] and |hash|.
+  1. If |hashExecResult| is null then return null.
+  1. Let |result| be the result of creating a new {{URLPatternResult}}.
```
domenic, line 323:
```suggestion
  1. Let |result| be a new {{URLPatternResult}}.
```

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Set |result|'s {{URLPatternResult/inputs}} to |inputs|.
+  1. Set |result|'s {{URLPatternResult/protocol}} to the result of [=create a component match result=] given |urlpattern|'s [=URLPattern/protocol component=], |protocol|, and |protocolExecResult|.
```
domenic, line 325:
```suggestion
  1. Set |result|'s {{URLPatternResult/protocol}} to the result of [=creating a component match result=] given |urlpattern|'s [=URLPattern/protocol component=], |protocol|, and |protocolExecResult|.
```

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Set |result|'s {{URLPatternInit/password}} to |password|.
+  1. Set |result|'s {{URLPatternInit/hostname}} to |hostname|.
+  1. Set |result|'s {{URLPatternInit/port}} to |port|.
+  1. Set |result|'s {{URLPatternInit/pathname}} to |pathname|.
+  1. Set |result|'s {{URLPatternInit/search}} to |search|.
+  1. Set |result|'s {{URLPatternInit/hash}} to |hash|.
+  1. If |init|'s {{URLPatternInit/baseURL}} is not null:
+    1. Let |baseURL| be the result of [=URL parser|parsing=] |init|'s {{URLPatternInit/baseURL}}.
+    1. If |baseURL| is failure, then throw a TypeError.
+    1. Set |result|'s {{URLPatternInit/protocol}} to |baseURL|'s [=URL/protocol].
```
domenic, line 351: Same problems here as above.

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+    1. Set |result|'s {{URLPatternInit/username}} to |baseURL|'s [=URL/username].
+    1. Set |result|'s {{URLPatternInit/password}} to |baseURL|'s [=URL/password].
+    1. Set |result|'s {{URLPatternInit/hostname}} to |baseURL|'s [=URL/hostname].
+    1. Set |result|'s {{URLPatternInit/port}} to |baseURL|'s [=URL/port].
+    1. Set |result|'s {{URLPatternInit/pathname}} to |baseURL|'s [=URL/pathname].
+    1. Set |result|'s {{URLPatternInit/search}} to |baseURL|'s [=URL/search].
+    1. Set |result|'s {{URLPatternInit/hash}} to |baseURL|'s [=URL/hash].
+  1. If |init|'s {{URLPatternInit/protocol}} is not null:
+    1. Set |result|'s {{URLPatternInit/protocol}} to the result of [=canonicalize protocol=] given |init|'s {{URLPatternInit/protocol}} and |type|.
```
domenic, line 360: IMO if you have a nested list with a single step, you should just collapse it into a single sentence.

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Return |result|.
+</div>
+
+<div algorithm>
+  To <dfn>create a component match result</dfn> given a [=component=] |component|, a string |input|, and an array representing the output of [=RegExpExec=] |execResult|:
+
+  1. Let |result| be the result of creating a new {{URLPatternComponentResult}}.
+  1. Set |result|'s {{URLPatternComponentResult/input}} to |input|.
+  1. Let |groups| be an empty object.
+  1. Let |index| = 1;
```
domenic, line 384:
```suggestion
  1. Let |index| be 1.
```

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. While |index| is less than |execResult|.length:
```
domenic, line 385: You can't do property access directly here. You need to use `[$Get$](|execResult|, "`length`")` with the following in your anchors block:

```
urlPrefix: https://tc39.es/ecma262/; spec: ECMASCRIPT
 type: abstract-op
  text: Get; url: #sec-get-o-p
```

Same below for the indexed properties. (Although there you'll also need to do `[$ToString$](|index|)` which will need a similar anchor.

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+    1. Let |name| be |component|'s [=component/group name list=][|index| - 1].
```
domenic, line 386:
```suggestion
    1. Let |name| be |component|'s [=component/group name list=][|index| &minus; 1].
```

looks slightly nicer :)

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+    1. Set |result|'s {{URLPatternInit/hash}} to the result of [=canonicalize hash=] given |init|'s {{URLPatternInit/hash}} and |type|.
+  1. Return |result|.
+</div>
+
+<div algorithm>
+  To <dfn>create a component match result</dfn> given a [=component=] |component|, a string |input|, and an array representing the output of [=RegExpExec=] |execResult|:
+
+  1. Let |result| be the result of creating a new {{URLPatternComponentResult}}.
+  1. Set |result|'s {{URLPatternComponentResult/input}} to |input|.
+  1. Let |groups| be an empty object.
```
domenic, line 383: I think it should be an empty `record<USVString, USVString>`

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Let |index| = 1;
+  1. While |index| is less than |execResult|.length:
+    1. Let |name| be |component|'s [=component/group name list=][|index| - 1].
+    1. Let |value| be |execResult|[|index|].
+    1. Set |groups|[|name|] to |value|.
+  1. Set |result|'s {{URLPatternComponentResult/groups}} to |groups|.
+  1. return |result|.
```
domenic, line 390:
```suggestion
  1. Return |result|.
```

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+<div algorithm>
+  To <dfn>canonicalize hash</dfn> given a string |value| and a string |type|:
+
+  1. TODO |value| |type|
+</div algorithm>
+
 <h2 id=components>Components</h2>
 
-<p>A {{URLPattern}} is associated with multiple <dfn>component</dfn> objects.
+A {{URLPattern}} is associated with multiple <dfn>component</dfn> objects.
```
domenic, line 443:
```suggestion
A {{URLPattern}} is associated with multiple <dfn>component</dfn> [=structs=].
```

spec.bs
```diff
@@ -243,8 +250,200 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+
+A [=component=] has an associated <dfn for=component>pattern string</dfn>, a string, initially null.
+
+A [=component=] has an associated <dfn for=component>regular expression</dfn>, a [=RegExp=], initially null.
```
domenic, line 447:
```suggestion
A [=component=] has an associated <dfn for=component>regular expression</dfn>, a {{RegExp}}, initially null.
```

spec.bs
```diff
@@ -18,6 +18,13 @@ Default Biblio Status: current
 Markup Shorthands: markdown yes
 </pre>
 
+<pre class="anchors">
+spec: ecma262; urlPrefix: https://tc39.es/ecma262/
+  type: dfn
```
domenic, line 23: You should sort these into their appropriate types. `RegExp` is an `interface`, and `RegExpExec` is an `abstract-op`. That will change how you link to them; you do `{{RegExp}}` and `[$RegExpExec$]`

domenic (commented):
This one was more exciting!

spec.bs
wanderview: For a lot of these I actually don't want the getter behavior anyway.  For example, I just want the scheme and not the trailing `:` added.  So I think I can reference the record values directly.  The only one I probably need to get exported is the pathname getter since it does extra work to construct the returned string with slashes, etc.

spec.bs
wanderview: I get an error with `{{RegExp}}`:

```
LINK ERROR: No 'idl' refs found for 'RegExp'.
{{RegExp}}
```

wanderview: Oh, I have to fix the anchors, nm.

wanderview:
Ok, I think I addressed all your feedback.  PTAL.

spec.bs
```diff
@@ -243,8 +258,189 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Let |protocol| be the empty string.
+  1. Let |username| be the empty string.
+  1. Let |password| be the empty string.
+  1. Let |hostname| be the empty string.
+  1. Let |port| be the empty string.
+  1. Let |pathname| be the empty string.
+  1. Let |search| be the empty string.
+  1. Let |hash| be the empty string.
+  1. Let |inputs| be an empty [=list=].
+  1. [=list/Append=] |input| on to |inputs|.
```
domenic, line 287:
```suggestion
  1. [=list/Append=] |input| to |inputs|.
```

spec.bs
```diff
@@ -243,8 +258,189 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+    1. Set |port| to |applyResult|["{{URLPatternInit/port}}"].
+    1. Set |pathname| to |applyResult|["{{URLPatternInit/pathname}}"].
+    1. Set |search| to |applyResult|["{{URLPatternInit/search}}"].
+    1. Set |hash| to |applyResult|["{{URLPatternInit/hash}}"].
+  1. Else:
+    1. Let |baseURL| be null.
+    1. If |baseURLString| was given, then:
+      1. Set |baseURL| to the result of [=URL parser|parsing=] |baseURLString|.
+      1. If |baseURL| is failure, return null.
+      1. Push |baseURL| on to |inputs|.
```
domenic, line 304:
```suggestion
      1. [=list/Append=] |baseURL| to |inputs|.
```

spec.bs
```diff
@@ -243,8 +258,189 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+    1. If |url| is failure, return null.
+    1. Set |protocol| to |url|'s [=url/scheme=].
+    1. Set |username| to |url|'s [=url/username=].
+    1. Set |password| to |url|'s [=url/password=].
+    1. Set |hostname| to |url|'s [=url/host=] or the empty string if the value is null.
+    1. Set |port| to |url|'s [=url/port=] or the empty string if the value is null.
+    1. Set |pathname| to |url|'s [=url/API pathname string=].
+    1. Set |search| to |url|'s [=url/query=] or the empty string if the value is null.
+    1. Set |hash| to |url|'s [=url/fragment=] or the empty string if the value is null.
+  1. Let |protocolExecResult| be the result of [$RegExpBuiltinExec$] given |urlpattern|'s [=URLPattern/protocol component=]'s [=component/regular expression=] and |protocol|.
```
domenic, line 315: You could optionally write these in function call form, i.e.
```suggestion
  1. Let |protocolExecResult| be [$RegExpBuiltinExec$](|urlpattern|'s [=URLPattern/protocol component=]'s [=component/regular expression=], |protocol|).
```

Up to you which you find clearer.

spec.bs
```diff
@@ -243,8 +258,189 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Set |result|["{{URLPatternInit/username}}"] to |username|.
+  1. Set |result|["{{URLPatternInit/password}}"] to |password|.
+  1. Set |result|["{{URLPatternInit/hostname}}"] to |hostname|.
+  1. Set |result|["{{URLPatternInit/port}}"] to |port|.
+  1. Set |result|["{{URLPatternInit/pathname}}"] to |pathname|.
+  1. Set |result|["{{URLPatternInit/search}}"] to |search|.
+  1. Set |result|["{{URLPatternInit/hash}}"] to |hash|.
+  1. If |init|["{{URLPatternInit/baseURL}}"] is not null:
+    1. Let |baseURL| be the result of [=URL parser|parsing=] |init|["{{URLPatternInit/baseURL}}"].
+    1. If |baseURL| is failure, then throw a {{TypeError}}.
```
domenic, line 351: This exception never surfaces to the user---its only call site catches it and converts it into null. This feels a bit strange. Will a future part of the spec use this algorithm directly in a way that the exception surfaces?

If not, I'd suggest returning a sentinel like null or failure.

spec.bs
```diff
@@ -243,8 +258,189 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. If |init|["{{URLPatternInit/hash}}"] is not null then set |result|["{{URLPatternInit/hash}}"] to the result of [=canonicalize hash=] given |init|["{{URLPatternInit/hash}}"] and |type|.
+  1. Return |result|.
+</div>
+
+<div algorithm>
+  To <dfn>create a component match result</dfn> given a [=component=] |component|, a string |input|, and an array representing the output of [$RegExpBuiltinExec$] |execResult|:
+
+  1. Let |result| be the result of creating a new {{URLPatternComponentResult}}.
+  1. Set |result|'s {{URLPatternComponentResult/input}} to |input|.
+  1. Let |groups| be a [=record=]<{{USVString}}, {{USVString}}>.
```
domenic, line 376:
```suggestion
  1. Let |groups| be a <code>[=record=]<{{USVString}}, {{USVString}}></code>.
```

I think this will work and will look very slightly nicer.

spec.bs
```diff
@@ -243,8 +258,189 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. If |init|["{{URLPatternInit/search}}"] is not null then set |result|["{{URLPatternInit/search}}"] to the result of [=canonicalize search=] given |init|["{{URLPatternInit/search}}"] and |type|.
+  1. If |init|["{{URLPatternInit/hash}}"] is not null then set |result|["{{URLPatternInit/hash}}"] to the result of [=canonicalize hash=] given |init|["{{URLPatternInit/hash}}"] and |type|.
+  1. Return |result|.
+</div>
+
+<div algorithm>
+  To <dfn>create a component match result</dfn> given a [=component=] |component|, a string |input|, and an array representing the output of [$RegExpBuiltinExec$] |execResult|:
+
+  1. Let |result| be the result of creating a new {{URLPatternComponentResult}}.
+  1. Set |result|'s {{URLPatternComponentResult/input}} to |input|.
```
domenic, line 375:
```suggestion
  1. Set |result|["{{URLPatternComponentResult/input}}"] to |input|.
```

spec.bs
```diff
@@ -243,8 +258,189 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. If |init|["{{URLPatternInit/pathname}}"] is not null then set |result|["{{URLPatternInit/pathname}}"] to the result of [=canonicalize pathname=] given |init|["{{URLPatternInit/pathname}}"] and |type|.
+  1. If |init|["{{URLPatternInit/search}}"] is not null then set |result|["{{URLPatternInit/search}}"] to the result of [=canonicalize search=] given |init|["{{URLPatternInit/search}}"] and |type|.
+  1. If |init|["{{URLPatternInit/hash}}"] is not null then set |result|["{{URLPatternInit/hash}}"] to the result of [=canonicalize hash=] given |init|["{{URLPatternInit/hash}}"] and |type|.
+  1. Return |result|.
+</div>
+
+<div algorithm>
+  To <dfn>create a component match result</dfn> given a [=component=] |component|, a string |input|, and an array representing the output of [$RegExpBuiltinExec$] |execResult|:
+
+  1. Let |result| be the result of creating a new {{URLPatternComponentResult}}.
```
domenic, line 374:
```suggestion
  1. Let |result| be a new {{URLPatternComponentResult}}.
```

spec.bs
```diff
@@ -243,8 +258,189 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Set |result|'s {{URLPatternComponentResult/input}} to |input|.
+  1. Let |groups| be a [=record=]<{{USVString}}, {{USVString}}>.
+  1. Let |index| be 1;
```
domenic, line 377:
```suggestion
  1. Let |index| be 1.
```

spec.bs
```diff
@@ -243,8 +258,189 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. While |index| is less than [$Get$](|execResult|, "length"):
```
domenic, line 378:
```suggestion
  1. While |index| is less than [$Get$](|execResult|, "`length`"):
```

spec.bs
```diff
@@ -243,8 +258,189 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+    1. Let |name| be |component|'s [=component/group name list=][|index| &minus; 1].
+    1. Let |value| be [$ToString$]([$Get$](|execResult|, |index|)).
```
domenic, line 380: You need to ToString() the index, not the result.

spec.bs
```diff
@@ -243,8 +258,189 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+    1. Set |groups|[|name|] to |value|.
+  1. Set |result|'s {{URLPatternComponentResult/groups}} to |groups|.
```
domenic, line 382:
```suggestion
  1. Set |result|["{{URLPatternComponentResult/groups}}"] to |groups|.
```

domenic (approved):
Looks great with nits

spec.bs
wanderview: Yes, the constructor will also call this and there errors will flow back to the caller.

spec.bs
wanderview: Uh, why do we need quotes within quotes?

PR #54, opened 2021-07-16
Attempt to canonicalize components for test()/exec().

wanderview:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/54.html" title="Last updated on Jul 20, 2021, 7:52 PM UTC (52d736e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/54/bd70f55...52d736e.html" title="Last updated on Jul 20, 2021, 7:52 PM UTC (52d736e)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -20,6 +20,7 @@ Markup Shorthands: markdown yes
 
 <pre class="link-defaults">
 spec:infra; type:dfn; text:list
+spec:url; type:dfn; for:/; text:url
 spec:webidl; type:dfn; text:record
 </pre>
 
@@ -31,6 +32,15 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: ToString; url: #sec-tostring
   type: interface
     text: RegExp; url: #sec-regexp-regular-expression-objects
+spec: url; urlPrefix: https://url.spec.whatwg.org/
+  type: dfn
+    text: cannot be a base URL path state; url: #cannot-be-a-base-url-path-state
+    text: fragment state; url: #fragment-state
+    text: hostname state; url: #hostname-state
+    text: path start state; url: #path-start-state
+    text: port state; url: #port-state
+    text: query state; url: #query-state
+    text: scheme start state; url: #scheme-start-state
 </pre>
 
 <style>
@@ -361,8 +371,8 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
   1. If |init|["{{URLPatternInit/username}}"] is not null then set |result|["{{URLPatternInit/username}}"] to the result of [=canonicalize username=] given |init|["{{URLPatternInit/username}}"] and |type|.
   1. If |init|["{{URLPatternInit/password}}"] is not null then set |result|["{{URLPatternInit/password}}"] to the result of [=canonicalize password=] given |init|["{{URLPatternInit/password}}"] and |type|.
   1. If |init|["{{URLPatternInit/hostname}}"] is not null then set |result|["{{URLPatternInit/hostname}}"] to the result of [=canonicalize hostname=] given |init|["{{URLPatternInit/hostname}}"] and |type|.
-  1. If |init|["{{URLPatternInit/port}}"] is not null then set |result|["{{URLPatternInit/port}}"] to the result of [=canonicalize port=] given |init|["{{URLPatternInit/port}}"] and |type|.
-  1. If |init|["{{URLPatternInit/pathname}}"] is not null then set |result|["{{URLPatternInit/pathname}}"] to the result of [=canonicalize pathname=] given |init|["{{URLPatternInit/pathname}}"] and |type|.
+  1. If |init|["{{URLPatternInit/port}}"] is not null then set |result|["{{URLPatternInit/port}}"] to the result of [=canonicalize port=] given |init|["{{URLPatternInit/port}}"], |result|["{{URLPatternInit/protocol}}"], and |type|.
+  1. If |init|["{{URLPatternInit/pathname}}"] is not null then set |result|["{{URLPatternInit/pathname}}"] to the result of [=canonicalize pathname=] given |init|["{{URLPatternInit/pathname}}"], |result|["{{URLPatternInit/protocol}}"], and |type|.
   1. If |init|["{{URLPatternInit/search}}"] is not null then set |result|["{{URLPatternInit/search}}"] to the result of [=canonicalize search=] given |init|["{{URLPatternInit/search}}"] and |type|.
   1. If |init|["{{URLPatternInit/hash}}"] is not null then set |result|["{{URLPatternInit/hash}}"] to the result of [=canonicalize hash=] given |init|["{{URLPatternInit/hash}}"] and |type|.
   1. Return |result|.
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic here is an attempt at canonicalizing the component values passed to `test()` and `exec()`.  I'm trying to use the URL parser infrastructure for this, but not sure if I got it right.  I wish I could run tests or something against the spec.

wanderview:
@domenic Ok.  The parent PR merged and I updated this a bit to be more correct.  Note:

* I purposely don't use state override for protocol since it prevents the scheme from being set in many cases.
* I forcibly set the scheme in port and pathname depending on it already having been canonicalized before getting passed in to those algorithms.

Not sure if there is a good place to note that kind of stuff in the spec text.

spec.bs
```diff
@@ -386,49 +396,91 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
 <div algorithm>
   To <dfn>canonicalize protocol</dfn> given a string |value| and a string |type|:
 
-  1. TODO |value| |type|
+  1. Let |strippedValue| be the given |value| with a single trailing U+003A (:) removed, if any.
+  1. If |type| is "pattern" then return |strippedValue|.
+  1. Let |dummyURL| be a new [=url=] record.
```
domenic, line 401:
```suggestion
  1. Let |dummyURL| be a new [=URL record=].
```

here and elsewhere

spec.bs
```diff
@@ -386,49 +396,91 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Let |parseResult| be the result of [=Basic URL parser|URL parsing=] |strippedValue|, followed by U+003A (:), with |dummyURL| as [=basic URL parser/url=].
```
domenic:
```suggestion
  1. Let |parseResult| be the result of running the [=basic URL parser=] given |strippedValue| followed by U+003A (:), with |dummyURL| as [=basic URL parser/url=].
```

this was pretty confusing as I thought you were intentionally going through the full parser.

spec.bs
domenic: It seems like the algorithms are OK with an empty URL record... a bit surprising, but it seems to work. You could double-check with whatwg-url if you wanted.

spec.bs
```diff
@@ -386,49 +396,91 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
 
 <div algorithm>
-  To <dfn>canonicalize port</dfn> given a string |value| and a string |type|:
-
-  1. TODO |value| |type|
+  To <dfn>canonicalize port</dfn> given a string |portValue|, a string |protocolValue|, and a string |type|:
+
+  1. If |type| is "pattern" then return |portValue|.
+  1. Let |dummyURL| be a new [=url=] record.
+  1. Set |dummyURL|'s [=url/scheme=] to |protocolValue|.
```
domenic, line 440: I think here and line 452 would be good places to insert such notes.

Mixing `<p class="note">` with the markdown lists can be a bit tricky; you have to match the indent levels.

```
1. Set blah
   <p class="note">blah</p>
1. Blah blah
```

spec.bs
```diff
@@ -386,49 +396,91 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Let |parseResult| be the result of [=Basic URL parser|URL parsing=] |value| with |dummyURL| as [=basic URL parser/url=] and [=hostname state=] as [=basic URL parser/state override=].
+  1. If |parseResult| is failure, then throw a {{TypeError}}.
+  1. Return |dummyURL|'s [=url/host=].
 </div algorithm>
 
 <div algorithm>
-  To <dfn>canonicalize port</dfn> given a string |value| and a string |type|:
-
-  1. TODO |value| |type|
+  To <dfn>canonicalize port</dfn> given a string |portValue|, a string |protocolValue|, and a string |type|:
```
domenic, line 436:
```suggestion
  To <dfn>canonicalize port</dfn> given a string |portValue|, a string-or-null |protocolValue|, and a string |type|:
```

domenic: This algorithm's current call site only gives one argument. I think it should explicitly give null for the second.

spec.bs
```diff
@@ -386,49 +396,91 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
-
-  1. TODO |value| |type|
+  To <dfn>canonicalize pathname</dfn> given a string |pathnameValue|, a string |protocolValue|, and a string |type|:
+
+  1. If |type| is "pattern" then return |pathnameValue|.
+  1. Let |dummyURL| be a new [=url=] record.
+  1. If |protocolValue| is null or the empty string, then set |protocolValue| to "https".
+  1. Set |dummyURL|'s [=url/scheme=] to |protocolValue|.
+  1. If |dummyURL| [=is special=], then:
+    1. Let |parseResult| be the result of [=Basic URL parser|URL parsing=] |pathnameValue| with |dummyURL| as [=basic URL parser/url=] and [=path start state=] as [=basic URL parser/state override=].
```
domenic, line 454: It's a bit confusing to use two variables both named parseResult for different purposes in these cases. I'd name this one differently.

spec.bs
```diff
@@ -386,49 +396,91 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Let |parseResult| be the result of [=Basic URL parser|URL parsing=] |portValue| with |dummyURL| as [=basic URL parser/url=] and [=port state=] as [=basic URL parser/state override=].
+  1. If |parseResult| is failure, then throw a {{TypeError}}.
+  1. Return |dummyURL|'s [=url/port=].
 </div algorithm>
 
 <div algorithm>
-  To <dfn>canonicalize pathname</dfn> given a string |value| and a string |type|:
-
-  1. TODO |value| |type|
+  To <dfn>canonicalize pathname</dfn> given a string |pathnameValue|, a string |protocolValue|, and a string |type|:
```
domenic, line 447: Similar call- site mismatch.

spec.bs
```diff
@@ -386,49 +396,91 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
 <div algorithm>
   To <dfn>canonicalize protocol</dfn> given a string |value| and a string |type|:
 
-  1. TODO |value| |type|
+  1. Let |strippedValue| be the given |value| with a single trailing U+003A (:) removed, if any.
+  1. If |type| is "pattern" then return |strippedValue|.
```
domenic, line 400:
```suggestion
  1. If |type| is "`pattern`" then return |strippedValue|.
```

here and elsewhere

wanderview:
Sorry, had to push another change for `canonicalizeProtocol()`.  I did some testing with whatwg-url and realized that I needed to provide `://` and a dummy hostname in order for special schemes to be accepted.  Kind of ugly, but I think its the only way to do this.

spec.bs
domenic: Here is a good place to insert a note about why using state override doesn't work.

spec.bs
wanderview: This was a mistake.  It should be passing a protocol.

wanderview:
@domenic I think I address the feedback.  I also tested each component in whatwg-url.  I fixed another problem I with cannot-be-a-base-URL pathnames where I had to init the property to empty string.

Also, it seems the bikeshed server is having trouble...

spec.bs
```diff
@@ -386,49 +396,95 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
 <div algorithm>
   To <dfn>canonicalize protocol</dfn> given a string |value| and a string |type|:
 
-  1. TODO |value| |type|
+  1. Let |strippedValue| be the given |value| with a single trailing U+003A (:) removed, if any.
+  1. If |type| is "`pattern`" then return |strippedValue|.
+  1. Let |dummyURL| be a new [=url record=].
```
domenic, line 401:
```suggestion
  1. Let |dummyURL| be a new [=URL record=].
```

here and throughout

spec.bs
```diff
@@ -386,49 +396,95 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Let |parseResult| be the result of running the [=basic URL parser=] given |strippedValue| followed by "`://dummy.test`", with |dummyURL| as [=basic URL parser/url=].
```
domenic, line 402:
```suggestion
  1. Let |parseResult| be the result of running the [=basic URL parser=] given |strippedValue| followed by "`://dummy.test`", with |dummyURL| as <i>[=basic URL parser/url=]</i>.
```

Also for `[=basic URL parser/state override=]`. See https://infra.spec.whatwg.org/#algorithm-params and some background discussion in https://github.com/whatwg/infra/pull/325#issuecomment-702373931

spec.bs
```diff
@@ -386,49 +396,95 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. If |type| is "`pattern`" then return |pathnameValue|.
+  1. Let |dummyURL| be a new [=url record=].
+  1. If |protocolValue| is null or the empty string, then set |protocolValue| to "`https`".
+  1. Set |dummyURL|'s [=url/scheme=] to |protocolValue|.
+     <p class="note">Note, we set the [=url record=]'s [=url/scheme=] in order to check if it [=is special=].  If |protocolValue| is missing we purposely default the URL to being special by using "`https`" for the [=url/scheme=].</p>
+  1. Let |parseResult| be null.
+  1. If |dummyURL| [=is special=], then:
+    1. Set |parseResult| to the result of running [=basic URL parser=] given |pathnameValue| with |dummyURL| as [=basic URL parser/url=] and [=path start state=] as [=basic URL parser/state override=].
+  1. Else:
+    1. Set |dummyURL|'s [=url/path=][`0`] to empty string.
```
domenic, line 460:
```suggestion
    1. Set |dummyURL|'s [=url/path=][0] to empty string.
```

for some reason we typically don't code-ify numbers :)

domenic (approved):
Looks great with only typographical nits

PR #56, opened 2021-07-20
Spec the pattern string tokenizer.

wanderview:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/56.html" title="Last updated on Jul 22, 2021, 5:01 PM UTC (70b4308)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/56/190f624...70b4308.html" title="Last updated on Jul 22, 2021, 5:01 PM UTC (70b4308)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -30,6 +30,9 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
     text: Get; url: #sec-get-o-p
     text: RegExpBuiltinExec; url: #sec-regexpbuiltinexec
     text: ToString; url: #sec-tostring
+  type: dfn
+    text: IdentifierPart; url: #prod-IdentifierPart
+    text: IdentifierStart; url: #prod-IdentifierStart
   type: interface
     text: RegExp; url: #sec-regexp-regular-expression-objects
 spec: url; urlPrefix: https://url.spec.whatwg.org/
@@ -396,7 +399,7 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
 <div algorithm>
   To <dfn>canonicalize protocol</dfn> given a string |value| and a string |type|:
 
-  1. Let |strippedValue| be the given |value| with a single trailing U+003A (:) removed, if any.
+  1. Let |strippedValue| be the given |value| with a single trailing U+003A (`:`) removed, if any.
   1. If |type| is "`pattern`" then return |strippedValue|.
   1. Let |dummyURL| be a new [=URL record=].
   1. Let |parseResult| be the result of running the [=basic URL parser=] given |strippedValue| followed by "`://dummy.test`", with |dummyURL| as <i>[=basic URL parser/url=]</i>.
@@ -466,7 +469,7 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
 <div algorithm>
   To <dfn>canonicalize search</dfn> given a string |value| and a string |type|:
 
-  1. Let |strippedValue| be the given |value| with a single leading U+003F (?) removed, if any.
+  1. Let |strippedValue| be the given |value| with a single leading U+003F (`?`) removed, if any.
   1. If |type| is "`pattern`" then return |strippedValue|.
   1. Let |dummyURL| be a new [=URL record=].
   1. Set |dummyURL|'s [=url/query=] to the empty string.
@@ -478,7 +481,7 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
 <div algorithm>
   To <dfn>canonicalize hash</dfn> given a string |value| and a string |type|:
 
-  1. Let |strippedValue| be the given |value| with a single leading U+0023 (#) removed, if any.
+  1. Let |strippedValue| be the given |value| with a single leading U+0023 (`#`) removed, if any.
   1. If |type| is "`pattern`" then return |strippedValue|.
   1. Let |dummyURL| be a new [=URL record=].
   1. Set |dummyURL|'s [=url/fragment=] to the empty string.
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic PTAL.  This is a bigger pull request that specs the lexical tokenizer from the c++ implementation here:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/liburlpattern/tokenize.cc

spec.bs
```diff
@@ -491,12 +494,243 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+
+<h3 id=parsing-patterns>Parsing Patterns</h3>
+
+<h4 id=tokens>Tokens</h4>
+
+A <dfn>token list</dfn> is a [=list=] containing zero or more [=token=] [=structs=].
+
+A <dfn>token</dfn> is a [=struct=] representing a single lexical token within a [=/pattern string=].
+
+A [=token=] has an associated <dfn for=token>type</dfn>, a string, initially "[=token/type/invalid-char=]".  It must be one of the following:
```
domenic, line 515: Linking to enum values while getting them to be in code font is annoying in Bikeshed. The best I've found is `"<a for="token/type"><code>invalid-char</code></a>"`. I'm not sure it's worth it...

spec.bs
```diff
@@ -396,7 +399,7 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
 <div algorithm>
   To <dfn>canonicalize protocol</dfn> given a string |value| and a string |type|:
 
-  1. Let |strippedValue| be the given |value| with a single trailing U+003A (:) removed, if any.
+  1. Let |strippedValue| be the given |value| with a single trailing U+003A (`:`) removed, if any.
```
domenic, line 402: Making these be in code font is not typical but it's a fine style choice if you find it cleaner :)

spec.bs
```diff
@@ -491,12 +494,243 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+
+  1. Let |token| be a new [=token=].
+  1. Set |token|'s [=token/type=] to |type|.
+  1. Set |token|'s [=token/index=] to |tokenizer|'s [=tokenizer/index=].
+  1. Set |token|'s [=token/value=] to the substring within |tokenizer|'s [=tokenizer/input=] starting at |value position| with |value length|.
+  1. Set |tokenizer|'s [=token/index=] to |next position|.
+</div>
+
+<div algorthm>
+  To <dfn>add a token with default length</dfn> for a given [=tokenizer=] |tokenizer|, [=token/type=] |type|, number |next position|, and number |value position|.
```
domenic, line 587:
```suggestion
  To <dfn>add a token with default length</dfn> for a given [=tokenizer=] |tokenizer|, [=token/type=] |type|, number |next position|, and number |value position|:
```

spec.bs
```diff
@@ -491,12 +494,243 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+</div>
+
+<div algorithm>
+  To <dfn>process a tokenizing error with default position</dfn> for a given [=tokenizer=] |tokenizer|:
+
+  1. Run [=process a tokenizing error=] given |tokenizer|, |tokenizer|'s [=tokenizer/next index=], and |tokenizer|'s [=tokenizer/index=].
+</div>
+
+<div algorithm>
+  To perform <dfn>is a valid name code point</dfn> given a unicode |code point| and a boolean |first|:
```
domenic, line 614:
```suggestion
  To perform <dfn>is a valid name code point</dfn> given a Unicode |code point| and a boolean |first|:
```

here and elsewhere

spec.bs
```diff
@@ -491,12 +494,243 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+<div algorithm>
+  To <dfn>process a tokenizing error</dfn> for a given [=tokenizer=] |tokenizer|, a number |next position|, and a number |value position|:
+
+  1. If |tokenizer|'s [=tokenizer/policy=] is "[=tokenize policy/strict=]", then throw a {{TypeError}}.
+  1. [=Assert=]: |tokenizer|'s [=tokenizer/policy=] is "[=tokenize policy/lenient=]".
+  1. Run [=add a token with default length=] given |tokenizer|, "[=token/type/invalid-char=]", |next position|, and |value position|.
+</div>
+
+<div algorithm>
+  To <dfn>process a tokenizing error with default position</dfn> for a given [=tokenizer=] |tokenizer|:
```
domenic, line 608: This and "is a valid name code point" only have one call site. It might be clearer to inline them, or it might be nice to keep them as semantic helpers.

spec.bs
```diff
@@ -491,12 +494,243 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Otherwise return false.
+</div>
+
+<div algorithm>
+  To <dfn export>tokenize</dfn> a given [=/pattern string=] |input| and [=tokenize policy=] |policy|:
+
+  1. Let |tokenizer| be a new [=tokenizer=].
+  1. Set |tokenizer|'s [=tokenizer/input=] to |input|.
+  1. Set |tokenizer|'s [=tokenizer/policy=] to |policy|.
+  1. Let |input length| be [$Get$](|input|, "`length`").
```
domenic, line 633: You are not working with JS objects here so you should just say `|input|'s [=string/length=]`.

spec.bs
```diff
@@ -491,12 +494,243 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+      1. Let |name position| be |tokenizer|'s [=tokenizer/next index=].
+      1. Let |name start| be |name position|.
+      1. While |name position| is less than |input length|:
+        1. Run [=seek and get the next code point=] given |tokenizer| and |name position|.
+        1. Let |first code point| be true if |name position| equals |name start| and false otherwise.
+        1. Let |valid code point| be the result of running [=is a valid name code point=] given |tokenizer|'s [=tokenizer/code point=] and |first code point|.
+        1. If |valid code point| is false [=break=].
+        1. Set |name position| to |tokenizer|'s [=tokenizer/next index=].
+      1. If |name position| is less than or equal to |name start|:
+        1. Run [=process a tokenizing error=] given |tokenizer|, |name start|, and |tokenizer's| [=tokenizer/index=].
```
domenic, line 667:
```suggestion
        1. Run [=process a tokenizing error=] given |tokenizer|, |name start|, and |tokenizer|'s [=tokenizer/index=].
```

spec.bs
```diff
@@ -491,12 +494,243 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+        1. Run [=seek and get the next code point=] given |tokenizer| and |regex position|.
+        1. If the result of running [=is ASCII=] given |tokenizer|'s [=tokenizer/code point=] is false:
+          1. Run [=process a tokenizing error=] given |tokenizer|, |regex start|, and |tokenizer|'s [=tokenizer/index=].
+          1. Set |error| to true.
+          1. [=Break=].
+        1. If |regex position| equals |regex start| and |tokenizer|'s [=tokenizer/code point=] is U+003F (`?`):
+          1. Run [=process a tokenizing error=] given |tokenizer|, |regex start|, and |tokenizer|'s [=tokenizer/index=].
+          1. Set |error| to true.
+          1. [=Break=].
+        1. If |tokenizer|'s [=tokenizer/code point=] is U+005C (&bpos;):
```
domenic, line 685: I think this should be `\` not `&bpos;`?

spec.bs
```diff
@@ -491,12 +494,243 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
-A [=component=] has an associated <dfn for=component>pattern string</dfn>, a string, initially null.
+A [=component=] has an associated <dfn for=component>pattern string</dfn>, a [=/pattern string=], initially null.
 
 A [=component=] has an associated <dfn for=component>regular expression</dfn>, a {{RegExp}}, initially null.
 
 A [=component=] has an associated <dfn for=component>group name list</dfn>, an array of strings, initially an empty array.
 
+<h2 id=patterns>Patterns</h2>
+
+A <dfn>pattern string</dfn> is a string conforming to a particular syntax.  This pattern syntax is directly based on the syntax used by the popular [path-to-regexp](https://github.com/pillarjs/path-to-regexp) JavaScript library.
```
domenic, line 505: Is this definition accurate? In particular it seems like the tokenizers use of pattern strings as inputs does not assume they conform to any particular syntax; it tokenizes anything you give it.

Maybe something like

> A pattern string is a string that will be processed as part of URL pattern tokenizing and parsing

?

wanderview: Hmm.  I'll rephrase to say that a "well formed" or "valid" pattern string conforms to a particular syntax.  And I can change the input of the tokenizer to not be a `[=pattern string=]`.

spec.bs
wanderview: Bikeshed formatting broke when I tried to use `\`.  Not sure how to put a backslash literal in.

spec.bs
wanderview: I personally like having them separate to help keep the main algorithm as clean as possible.  Maybe we could inline the one about default position.  I thought I used that more.

spec.bs
wanderview: I switched to `<code>\</code>` which seems to work.

spec.bs
wanderview: Switched them over using backticks instead of `<code>`.

spec.bs
wanderview: Ok.  I inlined the "process a tokenizing error with default position" call.  Keeping the "is a valid name code point" because I think the algorithm name is informative of the purpose.

spec.bs
wanderview: Rephrased, although, I still not 100% sure about this yet.  Might need to redo it once the rest is written.

wanderview:
@domenic PTAL

spec.bs
```diff
@@ -491,12 +494,236 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+
+<h4 id=tokenizing>Tokenizing</h4>
+
+A <dfn>tokenize policy</dfn> is a string that must be either "<dfn for="tokenize policy">`strict`</dfn>" or "<dfn for="tokenize policy">`lenient`</dfn>".
+
+A <dfn>tokenizer</dfn> is a [=struct=].
+
+A [=tokenizer=] has an associated <dfn for=tokenizer>input</dfn>, a [=/pattern string=], initially the empty string.
+
+A [=tokenizer=] has an associated <dfn for=tokenizer>policy</dfn>, a [=tokenize policy=], initially "[=tokenize policy/strict=]".
```
domenic, line 552: Here and in two other instances (one strict one lenient) you can make these code font.

spec.bs
```diff
@@ -491,12 +494,236 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+A <dfn>token</dfn> is a [=struct=] representing a single lexical token within a [=/pattern string=].
+
+A [=token=] has an associated <dfn for=token>type</dfn>, a string, initially "<a for=token/type>`invalid-char`</a>".  It must be one of the following:
+
+<dl>
+  <dt>"<dfn for=token/type>`open`</dfn>"</dt>
+  <dd>The [=token=] represents a U+007B (`{`) code point.
+  <dt>"<dfn for=token/type>`close`</dfn>"</dt>
+  <dd>The [=token=] represents a U+007D (`}`) code point.
+  <dt>"<dfn for=token/type>`regex`</dfn>"</dt>
```
domenic, line 522: Oh, I forgot this one last time: it's traditional to abbreviate regular expressions as "regexp", not "regex", in the web ecosystem.

spec.bs
```diff
@@ -491,12 +494,236 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+      1. Run [=add a token with default position and length=] given |tokenizer| and "<a for=token/type>`other-modifier`</a>".
+      1. [=Continue=].
+    1. If |tokenizer|'s [=tokenizer/code point=] is U+005C (<code>\</code>):
+      1. If |tokenizer|'s [=tokenizer/index=] is equal to |tokenizer|'s [=tokenizer/input=]'s [=string/length=] &minus; 1:
+        1. Run [=process a tokenizing error=] given |tokenizer|, |tokenizer|'s [=tokenizer/next index=], and |tokenizer|'s [=tokenizer/index=].
+        1. [=Continue=].
+      1. Let |escaped index| be |tokenizer|'s [=tokenizer/next index=].
+      1. Run [=get the next code point=] given |tokenizer|.
+      1. Run [=add a token with default length=] given |tokenizer|, "<a for=token/type>`escaped-char`</a>", |tokenizer|'s [=tokenizer/next index=], and |escaped index|.
+
```
domenic, line 642: Extra empty line?

domenic (approved):
LGTM with nits

spec.bs
domenic: It looks pretty good for now. Ideally you'd have a rigorous definition of well-formed that can be used by validators, but that's probably overkill. Maybe "well-formed" really means "already validated by the parser/tokenizer" or something in your case?

Anyway, yeah, leaving the exact distinction here for a polish phase makes sense.

PR #57, opened 2021-07-21
Spec the pattern string parser.

wanderview:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/57.html" title="Last updated on Jul 29, 2021, 6:27 PM UTC (7a58f60)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/57/8cb3403...7a58f60.html" title="Last updated on Jul 29, 2021, 6:27 PM UTC (7a58f60)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -560,66 +560,7 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>next index</dfn>, a number,
 A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode code point, initially null.
 
 <div algorithm>
-  To <dfn>get the next code point</dfn> for a given [=tokenizer=] |tokenizer|:
-
-  1. Set |tokenizer|'s [=tokenizer/code point=] to the Unicode code point in |tokenizer|'s [=tokenizer/input=] at the position indicated by |tokenizer|'s [=tokenizer/next index=].
-  1. Increment |tokenizer|'s [=tokenizer/next index=] by 1.
-</div>
-
-<div algorithm>
-  To <dfn>seek and get the next code point</dfn> for a given [=tokenizer=] |tokenizer| and number |index|:
-
-  1. Set |tokenizer|'s [=tokenizer/next index=] to |index|.
-  1. Run [=get the next code point=] given |tokenizer|.
-</div>
-
-<div algorithm>
-  To <dfn>add a token</dfn> for a given [=tokenizer=] |tokenizer|, [=token/type=] |type|, number |next position|, number |value position|, and number |value length|:
-
-  1. Let |token| be a new [=token=].
-  1. Set |token|'s [=token/type=] to |type|.
-  1. Set |token|'s [=token/index=] to |tokenizer|'s [=tokenizer/index=].
-  1. Set |token|'s [=token/value=] to the substring within |tokenizer|'s [=tokenizer/input=] starting at |value position| with |value length|.
-  1. Set |tokenizer|'s [=token/index=] to |next position|.
-</div>
-
-<div algorthm>
-  To <dfn>add a token with default length</dfn> for a given [=tokenizer=] |tokenizer|, [=token/type=] |type|, number |next position|, and number |value position|:
-
-  1. Let |computed length| be |next position| &minus; |value position|.
-  1. Run [=add a token=] given |tokenizer|, |type|, |next position|, |value position|, and |computed length|.
-</div>
-
-<div algorithm>
-  To <dfn>add a token with default position and length</dfn> for a given [=tokenizer=] |tokenizer| and [=token/type=] |type|:
-
-  1. Run [=add a token with default length=] given |tokenizer|, |type|, |tokenizer|'s [=tokenizer/next index=], and |tokenizer|'s [=tokenizer/index=].
-</div>
-
-<div algorithm>
-  To <dfn>process a tokenizing error</dfn> for a given [=tokenizer=] |tokenizer|, a number |next position|, and a number |value position|:
-
-  1. If |tokenizer|'s [=tokenizer/policy=] is "<a for="tokenize policy">`strict`</a>", then throw a {{TypeError}}.
-  1. [=Assert=]: |tokenizer|'s [=tokenizer/policy=] is "<a for="tokenize policy">`lenient`</a>".
-  1. Run [=add a token with default length=] given |tokenizer|, "<a for=token/type>`invalid-char`</a>", |next position|, and |value position|.
-</div>
-
-<div algorithm>
-  To perform <dfn>is a valid name code point</dfn> given a Unicode |code point| and a boolean |first|:
-
-  1. If |first| is true return the result of checking if |code point| is contained in the [=IdentifierStart=] set of code points.
-  1. Else return the result of checking if |code point| is contained in the [=IdentifierPart=] set of code points.
-</div>
-
-<div algorithm>
-  To determine if a Unicode |code point| <dfn>is ASCII</dfn>:
-
-  1. If |code point| is between U+0000 and U+007F inclusive, then return true.
-  1. Otherwise return false.
-</div>
-
-<div algorithm>
-  To <dfn export>tokenize</dfn> a given string |input| and [=tokenize policy=] |policy|:
+  To <dfn>tokenize</dfn> a given string |input| and [=tokenize policy=] |policy|:
 
   1. Let |tokenizer| be a new [=tokenizer=].
   1. Set |tokenizer|'s [=tokenizer/input=] to |input|.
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic PTAL.  The c++ source code for this PR can be found here:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/liburlpattern/parse.cc

spec.bs
```diff
@@ -719,6 +719,233 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
   1. Return |tokenizer|'s [=tokenizer/token list=].
 </div>
 
+<h4 id=parts>Parts</h4>
+
+A <dfn>part list</dfn> is a [=list=] of zero or more [=part=] [=structs=].
```
domenic, line 724:
```suggestion
A <dfn>part list</dfn> is a [=list=] of zero or more [=parts=].
```

spec.bs
```diff
@@ -719,6 +719,233 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
...
+  <dd>The [=part=] has an optional [=part/modifier=] indicated by the U+003F (`?`) code point.</dd>
+  <dt>"<dfn for=part/modifier>`zero-or-more`</dfn>"</dt>
+  <dd>The [=part=] has a "zero or more" [=part/modifier=] indicated by the U+002A (`*`) code point.</dd>
+  <dt>"<dfn for=part/modifier>`one-or-more`</dfn>"</dt>
+  <dd>The [=part=] has a "one or more" [=part/modifier=] indicated by the U+002B (`+`) code point.</dd>
+</dl>
+
+<h4 id=options-header>Options</h4>
+
+An <dfn>options</dfn> [=struct=] contains different settings that control how [=/pattern string=] behaves.  These options originally come from [path-to-regexp](https://github.com/pillarjs/path-to-regexp).  Not all are utilized when constructing a {{URLPattern}}, but they are maintained here to make it easier to stay aligned with [path-to-regexp](https://github.com/pillarjs/path-to-regexp).
```
domenic, line 764: Sorry, but you should not include in the spec behavior that is not observable.

spec.bs
```diff
@@ -719,6 +719,233 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
...
+An <dfn>encoding callback</dfn> is an abstract algorithm that takes a given string |input|.  The |input| will be a simple text piece of a [=/pattern string=].  An implementing algorithm should validate and encode the |input|.  It must return the encoded string or throw an exception.
+</div>
+
+A <dfn>pattern parser</dfn> is a [=struct=].
+
+A [=pattern parser=] has an associated <dfn for="pattern parser">token list</dfn>, a [=/token list=], initially an empty [=list=].
+
+A [=pattern parser=] has an associated <dfn for="pattern parser">encoding callback</dfn>, a [=/encoding callback=], initially null.
+
+A [=pattern parser=] has an associated <dfn for="pattern parser">options</dfn>, a [=/options=], initially null.
```
domenic, line 790: Optional, but perhaps a bit clearer: if there are mandatory fields that are always set upon creation, you can say "which must be set upon creation". Then when creating you can say something like "a new pattern parser whose encoding callback is X and options is Y".

This makes it obvious which of these defaults are real defaults that are going to stick around. Right now it seems like some of these defaults are arbitrary and could be anything because you always overwrite them.

spec.bs
```diff
@@ -719,6 +719,233 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
...
+
+A [=pattern parser=] has an associated <dfn for="pattern parser">pending fixed value</dfn>, a string, initially the empty string.
+
+A [=pattern parser=] has an associated <dfn for="pattern parser">index</dfn>, a number, initially 0.
+
+A [=pattern parser=] has an associated <dfn for="pattern parser">next numeric name</dfn>, a number, initially 0.
+
+The <dfn>full wildcard regexp value</dfn> is the string "`.*`".
+
+<div algorithm>
```
domenic, line 804: It's a style thing, but I prefer putting the most important algorithms (in this case "parse a pattern") at the top of the section, and putting auxiliary helpers afterward.

spec.bs
```diff
@@ -719,6 +719,233 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
...
+To <dfn>try to consume a token</dfn> given a [=pattern parser=] |parser| and [=token/type=] |type|:
+
+  1. Let |next token| be |parser|'s [=pattern parser/token list=][|parser|'s [=pattern parser/index=]].
```
domenic, line 807: It's not clear from what I've read so far that the index will always point to a token that exists. Indeed I guess it's not true in the initial state (empty list/index 0).

It might be good to either assert that, if it's true for all callers, or add a note after the definition of index saying that that's always true once some initial setup is performed.

spec.bs
```diff
@@ -719,6 +719,233 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
...
+  1. If |result| is null, then throw a {{TypeError}}.
+  1. Return |result|.
+</div>
+
+<div algorithm>
+To <dfn>consume text</dfn> given a [=pattern parser=] |parser|:
+
+  1. Let |result| be the empty string.
+  1. Let |done| be false.
+  1. While |done| is false:
```
domenic, line 835: Can you avoid the done variable by just using `[=iteration/break=]` and "While true:"?

spec.bs
domenic: Same goes for the part struct it seems.

spec.bs
```diff
@@ -719,6 +719,233 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
...
+To <dfn>add a part</dfn> given a [=pattern parser=] |parser|, a string |prefix|, a [=token=] |name token|, a [=token=] |regexp or wildcard token|, a string |suffix|, and a [=token=] |modifier token|:
+
+  1. Let |modifier| be "<a for=part/modifier>`none`</a>".
+  1. If |modifier token| is not null:
+    1. If |modifier token|'s [=token/value=] is "`?`" then set |modifier| to "<a for=part/modifier>`optional`</a>".
+    1. Else if |modifier token|'s [=token/value=] is "`*`" then set |modifier| to "<a for=part/modifier>`zero-or-more`</a>".
+    1. Else if |modifier token|'s [=token/value=] is "`+`" then set |modifier| to "<a for=part/modifier>`one-or-more`</a>".
+  1. If |name token| is null and |regexp or wildcard token| is null and |modifier| is "<a for=part/modifier>`none`</a>":
+    1. Append |prefix| to the end of |parser|'s [=pattern parser/pending fixed value=].
+    1. Return
```
domenic, line 866:
```suggestion
    1. Return.
```

spec.bs
```diff
@@ -719,6 +719,233 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
...
+  1. Run [=maybe add a part from the pending fixed value=] given |parser|.
+  1. Let |part| be a new [=part=].
+  1. If |name token| is null and |regexp or wildcard token| is null:
+    1. [=Assert=] |suffix| is the empty string.
```
domenic, line 870:
```suggestion
    1. [=Assert=]: |suffix| is the empty string.
```

spec.bs
```diff
@@ -719,6 +719,233 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
...
+  1. Let |parser| be a new [=pattern parser=].
+  1. Set |parser|'s [=pattern parser/options=] to |options|.
+  1. Set |parser|'s [=pattern parser/encoding callback=] to |encoding callback|.
+  1. Set |parser|'s [=pattern parser/token list=] to the result of running [=tokenize=] given |input| and "<a for="tokenize policy">`strict`</a>".
+  1. While |parser|'s [=pattern parser/index=] is less than |parser|'s [=pattern parser/token list=]'s [=list/size=]:
+    1. Let |char token| be the result of running [=try to consume a token=] given |parser| and "<a for=token/type>`char`</a>".
+    1. Let |name token| be the result of running [=try to consume a token=] given |parser| and "<a for=token/type>`name`</a>".
+    1. Let |regexp or wildcard token| be the result of running [=try to consume a token=] given |parser| and "<a for=token/type>`regexp`</a>".
+    1. If |name token| is null and |regexp or wildcard token| is null:
+      1. Set |regexp or wildcard token| to the result of running [=try to consume a token=] given |parser| and "<a for=token/type>`asterisk`</a>".
```
domenic, line 917: Probably this doesn't need to be nested.

spec.bs
wanderview: I really want a `do {} while()` loop.  What is the best spec construct for that?

domenic: I don't understand why a do/while would be necessary here. If the first token ends up being null, you do nothing, right? So just a normal while loop would work.

wanderview: I don't think that's the case?

https://source.chromium.org/chromium/chromium/src/+/main:third_party/liburlpattern/parse.cc;l=79-87;drc=ccecd870311e5326a355715201778816c841e8f0

I guess I can just duplicate the two "try to consume" steps to get the token before the loop.  Just using `While true:` is probably easier, though.

domenic: I'm still confused. Here is my proposal:

While true:

1. Let token be the result of running try to consume a token given parser and "char".
1. If token is null, then set token to the result of running try to consume a token given parser and "escaped-char".
1. If token is null, then break.
1. Append token's value to the end of result.

Let's say both "char" and "escaped-char" will return null for the first token. Then:

- in your version we run steps 1, 2 which have side effects. 3 sets done to true. 4 does nothing since it's "else". Then we go back to the top of the loop where done is false, so we're done.

- in my version we run steps 1, 2. 3 breaks. So we're done.

domenic: (I guess it's not clear whether there are actually side effects... it seems like the index doesn't advance in that case.)

wanderview: Yea, your proposal works.  Its equivalent to the "do while" without the break statement.

spec.bs
wanderview: The option defaults actually come from path-to-regexp, but I can drop them since we don't use them.

spec.bs
wanderview: Done.  And I factored it out to a separate algorithm since we need to do these steps at least twice.

wanderview:
@domenic PTAL.  I addressed your action items.  I also added a bit about the `segment wildcard regexp` that I forgot before.  And I tried adding some notes based on the c++ comments.  Maybe some of them are not helpful enough.  Let me know what you think.  Thanks.

spec.bs
```diff
@@ -719,8 +660,341 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
...
+An <dfn>encoding callback</dfn> is an abstract algorithm that takes a given string |input|.  The |input| will be a simple text piece of a [=/pattern string=].  An implementing algorithm should validate and encode the |input|.  It must return the encoded string or throw an exception.
+</div>
+
+A <dfn>pattern parser</dfn> is a [=struct=].
+
+A [=pattern parser=] has an associated <dfn for="pattern parser">token list</dfn>, a [=/token list=], initially an empty [=list=].
+
+A [=pattern parser=] has an associated <dfn for="pattern parser">encoding callback</dfn>, a [=/encoding callback=], that must be set upon creation.
+
+A [=pattern parser=] has an associated <dfn for="pattern parser">options</dfn>, a [=/options=], that must be set upon creation.
```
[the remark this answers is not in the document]
domenic: It looks like this field is never referenced. Will it be? Or maybe it's just necessary in C++ where closures are a pain, but in the spec you didn't need it.

spec.bs
```diff
@@ -719,8 +660,341 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
...
+      1. If |char token| is not null then set |prefix| to |char token|'s [=token/value=].
+      1. If |prefix| is not contained in |options|'s [=options/prefix list=]:
+        1. Append |prefix| to the end of |parser|'s [=pattern parser/pending fixed value=].
+        1. Set |prefix| to the empty string.
+      1. Run [=maybe add a part from the pending fixed value=] given |parser|.
+      1. Let |modifier token| be the result of running [=try to consume a modifier token=] given |parser|.
+      1. Run [=add a part=] given |parser|, |prefix|, |name token|, |regexp or wildcard token|, the empty string, and |modifier token|.
+      1. [=Continue=].
+    1. Let |fixed token| be |char token|.
+      <p class=note>If there was no matching group, then we need to buffer an fixed text.  We want to collect as much text as possible before adding it as a "<a for=part/type>`fixed-text`</a>" [=part=].
```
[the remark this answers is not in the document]
wanderview: "any fixed text"

spec.bs
```diff
@@ -719,8 +660,341 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
...
+
+An [=/options=] has an associated <dfn for=options>delimiter list</dfn>, a string, which must be set upon creation.  Each code point in the [=options/delimiter list=] is treated as a segment separator.  This is used, for example, for determining how far a `:foo` named group should match by default.
+
+An [=/options=] has an associated <dfn for=options>prefix list</dfn>, a string, which must be set upon creation.  Each code point in the [=options/prefix list=] is treated as an automatic prefix if found immediately preceding a match group.  This matters when a match group is modified to be optional or repeating.  For example, if [=options/prefix list=] is "`/`" then "`/foo/:bar?/baz`" will treat the "`/`" before "`:bar`" as a prefix that becomes optional along with the named group.  So in this example the pattern would match "`/foo/baz`".
```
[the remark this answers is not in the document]
wanderview: We never use it with more than one code point, so I changed it to `prefix code point`.

spec.bs
```diff
@@ -719,8 +660,341 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
...
+  <dd>The [=part=] has a "zero or more" [=part/modifier=] indicated by the U+002A (`*`) code point.</dd>
+  <dt>"<dfn for=part/modifier>`one-or-more`</dfn>"</dt>
+  <dd>The [=part=] has a "one or more" [=part/modifier=] indicated by the U+002B (`+`) code point.</dd>
+</dl>
+
+<h4 id=options-header>Options</h4>
+
+An <dfn>options</dfn> [=struct=] contains different settings that control how [=/pattern string=] behaves.  These options originally come from [path-to-regexp](https://github.com/pillarjs/path-to-regexp).  We only include the options that are modified within the URLPattern specification and exclude the other options.  For the purposes of comparison, this specification acts like [path-to-regexp](https://github.com/pillarjs/path-to-regexp) where `sensitive`, `strict`, `start`, and `end` are always set to true.
+
+An [=/options=] has an associated <dfn for=options>delimiter list</dfn>, a string, which must be set upon creation.  Each code point in the [=options/delimiter list=] is treated as a segment separator.  This is used, for example, for determining how far a `:foo` named group should match by default.
```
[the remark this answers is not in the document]
wanderview: Rewrote this to be `delimiter code point` and added an example.

wanderview:
Updated.  @domenic PTAL.

domenic (approved):
Looks great; the examples and string vs. code point change is very helpful.

PR #64, opened 2021-07-29
Spec generating a pattern string from a part list.

wanderview:
@domenic Here is a PR to add pattern string generation.  It depends on the parser PR landing first, so it can wait until that is merged.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/64.html" title="Last updated on Jul 29, 2021, 7:14 PM UTC (2e0eae9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/64/642585d...2e0eae9.html" title="Last updated on Jul 29, 2021, 7:14 PM UTC (2e0eae9)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -995,6 +995,74 @@ To <dfn>escape a regexp string</dfn> given a string |input|:
 
 <h3 id=converting-part-lists-to-pattern-strings>Converting Part Lists to Pattern Strings</h3>
 
+<div algorithm>
+To <dfn export>generate a [=/pattern string=]</dfn> from a given [=/part list=] |part list| and [=/options=] |options|:
+
+  1. Let |result| be the empty string.
+  1. [=list/For each=] |part| of |part list|:
+    1. If |part|'s [=part/type=] is "<a for=part/type>`fixed-text`</a>" then:
+      1. If |part|'s [=part/modifier=] is "<a for=part/modifier>`none`</a>" then:
+        1. Append the result of running [=escape a pattern string=] given |part|'s [=part/value=] to the end of |result|.
+        1. [=Continue=].
+      1. Append "`{`" to the end of |result|.
+      1. Append the result of running [=escape a pattern string=] given |part|'s [=part/value=] to the end of |result|.
+      1. Append "`}`" to the end of |result|.
+      1. Append the result of running [=convert a modifier to a string=] given |part|'s [=part/modifier=] to the end of |result|.
+      1. [=Continue=].
+    1. Let |needs grouping| be true if at least one of the following are true, otherwise let it be false:
+      <ul>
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -995,6 +995,74 @@ To <dfn>escape a regexp string</dfn> given a string |input|:
...
+    1. If |part|'s [=part/type=] is "<a for=part/type>`fixed-text`</a>" then:
+      1. If |part|'s [=part/modifier=] is "<a for=part/modifier>`none`</a>" then:
+        1. Append the result of running [=escape a pattern string=] given |part|'s [=part/value=] to the end of |result|.
+        1. [=Continue=].
+      1. Append "`{`" to the end of |result|.
+      1. Append the result of running [=escape a pattern string=] given |part|'s [=part/value=] to the end of |result|.
+      1. Append "`}`" to the end of |result|.
+      1. Append the result of running [=convert a modifier to a string=] given |part|'s [=part/modifier=] to the end of |result|.
+      1. [=Continue=].
+    1. Let |needs grouping| be true if at least one of the following are true and otherwise false:
```
domenic, line 1012:
```suggestion
    1. Let |needs grouping| be true if at least one of the following are true; otherwise let it be false:
```

spec.bs
```diff
@@ -995,6 +995,74 @@ To <dfn>escape a regexp string</dfn> given a string |input|:
...
+        1. Append "`)`" to the end of |result|.
+      1. Else append "`*`" to the end of |result|.
+    1. Append the result of running [=escape a pattern string=] given |part|'s [=part/suffix=] to the end of |result|.
+    1. If |needs grouping| is true, then append "`}`" to the end of |result|.
+    1. Append the result of running [=convert a modifier to a string=] give |part|'s [=part/modifier=] to the end of |result|.
+  1. Return |result|.
+</div>
+
+<div algorithm>
+To <dfn>escape a pattern string</dfn> given a string |input|:
```
domenic, line 1045: You could unify this algorithm with "escape a regexp string" by having it take a list of escaped characters as an input, and defining the regexp escaped code points list and the pattern string escaped code points list. Similar to percent encoding/decoding and the various "percent-encode sets" in https://url.spec.whatwg.org/#percent-encoded-bytes .

I'd definitely do that if you have a third instance of this sort of algorithm. With only two it's less clear it would be worth it.

domenic (approved):

spec.bs
wanderview: I have them unified in the c++.  I guess I'd like to wait to unify them when I restructure and clean up the doc.  I think I'll end up creating a "utilities" section where they will be co-located and would make more sense then.  Going to leave as is for now, though.

PR #65, opened 2021-07-29
Spec regular expression generation.

wanderview:
@domenic PTAL

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/65.html" title="Last updated on Jul 29, 2021, 8:11 PM UTC (6b63868)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/65/dfa53fd...6b63868.html" title="Last updated on Jul 29, 2021, 8:11 PM UTC (6b63868)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -981,6 +981,77 @@ To <dfn>add a part</dfn> given a [=pattern parser=] |parser|, a string |prefix|,
 <h3 id=converting-part-lists-to-regular-expressions>Converting Part Lists to Regular Expressions</h3>
 
 <div algorithm>
+To <dfn export>generate a regular expression and name list</dfn> from a given [=/part list=] |part list| and [=/options=] |options|:
+
+  1. Let |result| be "`^`".
+  1. Let |name list| be a new [=list=].
+  1. [=list/For each=] |part| of |part list|:
+    1. If |part|'s [=part/type=] is "<a for=part/type>`fixed-text`</a>":
+      1. If |part|'s [=part/modifier=] is "<a for=part/modifier>`none`</a>", then append the result of running [=escape a regexp string=] given |part|'s [=part/value=] to the end of |result|.
+      1. Else:
+        <div class=note>
+        <p>A "<a for=part/type>`fixed-text`</a>" |part| with a modifier uses a non capturing group.  It uses the following form.
+        <p>`(?:<fixed text>)<modifier>`
+        </div>
+        1. Append "`(?:`" to the end of |result|.
+        1. Append the result of running [=escape a regexp string=] given |part|'s [=part/value=] to the end of |result|.
+        1. Append "`)`" to the end of |result|.
+        1. Append the result of running [=convert a modifier to a string=] give |part|'s [=part/modifier=] to the end of |result|.
+      1. [=Continue=].
+    1. [=Assert=]: |part|'s [=part/name=] is not the empty string and is not null.
+    1. [=list/Append=] |part|'s [=part/name=] to |name list|.
+      <p class=note>We collect the list of matching group names in a parallel list.  This is largely done for legacy reasons to match [path-to-regexp](https://github.com/pillarjs/path-to-regexp).  We could attempt to convert this to use regular expression named captured groups, but given the complexity of this algorithm there is a real risk of introducing unintended bugs.  In addition, if we ever end up exposing the generated regular expressions to the web we would like to maintain compability with [path-to-regexp](https://github.com/pillarjs/path-to-regexp) which has indicated its unlikely to switch to using named capture groups.
+    1. Let |regexp value| be |part|'s [=part/value=].
+    1. If |part|'s [=part/type=] is "<a for=part/type>`segment-wildcard`</a>", then set |regexp value| to the result of running [=generate a segment wildcard regexp=] give |options|.
+    1. Else if |part|'s [=part/type=] is "<a for=part/type>`full-wildcard`</a>", then set |regexp value| to [=full wildcard regexp value=].
+    1. If |part|'s [=part/prefix=] is the empty string and |part|'s [=part/suffix=] is the empty string:
+      <div class=note>
+      <p>If there is no [=part/prefix=] or [=part/suffix=] then we generate a simple capturing group.  It uses the following form.
+      <p>`(<regexp value>)<modifier>`
+      </div>
+      1. Append "`(`" to the end of |result|.
+      1. Append |regexp value| to the end of |result|.
+      1. Append "`)`" to the end of |result|.
+      1. Append the result of running [=convert a modifier to a string=] give |part|'s [=part/modifier=] to the end of |result|.
+      1. [=Continue=].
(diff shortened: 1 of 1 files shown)


wanderview:
c++ source is here:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/liburlpattern/pattern.cc;l=187;drc=73cd91d7a4f21a807709d8bed6b5eeb52cd521b4

You can see some representative regexp values in the tests:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/liburlpattern/pattern_unittest.cc;l=35;drc=02c1d179486e02c6c11fcc14031e494879a2ab5f

I didn't include those examples here because honestly they are unreadable.

wanderview:
Although a lot of those tests are using options we are not using in the spec.  So I guess we will have simpler regexp values.

spec.bs
```diff
@@ -980,6 +980,77 @@ To <dfn>add a part</dfn> given a [=pattern parser=] |parser|, a string |prefix|,
...
+      1. Append "`(?:`" to the end of |result|.
+      1. Append the result of running [=escape a regexp string=] given |part|'s [=part/prefix=] to the end of |result|.
+      1. Append "`(`" to the end of |result|.
+      1. Append |regexp value| to the end of |result|.
+      1. Append "`)`" to the end of |result|.
+      1. Append the result of running [=escape a regexp string=] given |part|'s [=part/suffix=] to the end of |result|.
+      1. Append "`)`" to the end of |result|.
+      1. Append the result of running [=convert a modifier to a string=] give |part|'s [=part/modifier=] to the end of |result|.
+      1. [=Continue=].
+    1. [=Assert=]: |part|'s [=part/modifier=] is "<a for=part/modifier>`zero-or-more`</a>" or "<a for=part/modifier>`one-or-more`</a>".
```
domenic, line 1031: Heh I see what you're doing here with the asserts.

I think it'd be a little more convention to have an "Else" block on this line and add the asserts indented under the else.

domenic (commented):
This isn't so bad... I like these small PRs.

spec.bs
wanderview: And drop the continue in the previous section?

wanderview: I accidentally indented things here.  I'll un-indent.

wanderview:
Updated. PTAL.

spec.bs
```diff
@@ -980,6 +980,77 @@ To <dfn>add a part</dfn> given a [=pattern parser=] |parser|, a string |prefix|,
...
+
+  1. Let |result| be "`^`".
+  1. Let |name list| be a new [=list=].
+  1. [=list/For each=] |part| of |part list|:
+    1. If |part|'s [=part/type=] is "<a for=part/type>`fixed-text`</a>":
+      1. If |part|'s [=part/modifier=] is "<a for=part/modifier>`none`</a>", then append the result of running [=escape a regexp string=] given |part|'s [=part/value=] to the end of |result|.
+      1. Else:
+        <div class=note>
+        <p>A "<a for=part/type>`fixed-text`</a>" |part| with a modifier uses a non capturing group with the form:
+        <p>`(?:<fixed text>)<modifier>`
```
domenic, line 994:
```suggestion
        <p>`(?:<fixed text>)<modifier>`.
```

spec.bs
```diff
@@ -980,6 +980,77 @@ To <dfn>add a part</dfn> given a [=pattern parser=] |parser|, a string |prefix|,
...
+      1. Append "`)`" to the end of |result|.
+      1. Append the result of running [=convert a modifier to a string=] give |part|'s [=part/modifier=] to the end of |result|.
+      1. [=Continue=].
+    1. [=Assert=]: |part|'s [=part/modifier=] is "<a for=part/modifier>`zero-or-more`</a>" or "<a for=part/modifier>`one-or-more`</a>".
+    1. [=Assert=]: |part|'s [=part/prefix=] is not the empty string or |part|'s [=part/suffix=] is not the empty string.
+      <div class=note>
+      <p class=allow-2119>Repeating parts with a [=part/prefix=] and/or [=part/suffix=] are dramatically more complicated.  We want to exclude the initial [=part/prefix=] and the final [=part/suffix=], but include them between any repeated elements.  To achieve this we provide a separate initial expression that excludes the [=part/prefix=].  Then the expression is duplicated with the [=part/prefix=]/[=part/suffix=] values included in an optional repeating element.  If zero values are permitted then a final optional modifier may be appended.  The overall form:
+      <p>`(?:<prefix>((?:<regexp value>)(?:<suffix><prefix>(?:<regexp value>))*)<suffix>)?`
```
domenic, line 1035:
```suggestion
      <p>`(?:<prefix>((?:<regexp value>)(?:<suffix><prefix>(?:<regexp value>))*)<suffix>)?`.
```

spec.bs
```diff
@@ -980,6 +980,77 @@ To <dfn>add a part</dfn> given a [=pattern parser=] |parser|, a string |prefix|,
...
+      1. Append the result of running [=escape a regexp string=] given |part|'s [=part/suffix=] to the end of |result|.
+      1. Append "`)`" to the end of |result|.
+      1. Append the result of running [=convert a modifier to a string=] give |part|'s [=part/modifier=] to the end of |result|.
+      1. [=Continue=].
+    1. [=Assert=]: |part|'s [=part/modifier=] is "<a for=part/modifier>`zero-or-more`</a>" or "<a for=part/modifier>`one-or-more`</a>".
+    1. [=Assert=]: |part|'s [=part/prefix=] is not the empty string or |part|'s [=part/suffix=] is not the empty string.
+      <div class=note>
+      <p class=allow-2119>Repeating parts with a [=part/prefix=] and/or [=part/suffix=] are dramatically more complicated.  We want to exclude the initial [=part/prefix=] and the final [=part/suffix=], but include them between any repeated elements.  To achieve this we provide a separate initial expression that excludes the [=part/prefix=].  Then the expression is duplicated with the [=part/prefix=]/[=part/suffix=] values included in an optional repeating element.  If zero values are permitted then a final optional modifier may be appended.  The overall form:
```
domenic, line 1034:
```suggestion
      <p class=allow-2119>Repeating parts with a [=part/prefix=] and/or [=part/suffix=] are dramatically more complicated.  We want to exclude the initial [=part/prefix=] and the final [=part/suffix=], but include them between any repeated elements.  To achieve this we provide a separate initial expression that excludes the [=part/prefix=].  Then the expression is duplicated with the [=part/prefix=]/[=part/suffix=] values included in an optional repeating element.  If zero values are permitted then a final optional modifier may be appended.  The overall form is:
```

domenic (approved):

spec.bs
wanderview: As discussed in chat, don't want to put a period at the end here since it might get confused as a regexp char.  I'll refactor to avoid this.

spec.bs
wanderview: Refactors this to "It uses the following form."

PR #66, opened 2021-07-29
Restructure doc and fill out canonicalization section.

wanderview:
***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Aug 10, 2021, 8:36 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2FWICG%2Furlpattern%2F83a6925ebe69c279bc9adb4aa14e02a9cfac5559%2Fspec.bs&md-warning=not%20ready)

```
Error running preprocessor, returned code: 1.
Traceback (most recent call last):
  File "/sites/api.csswg.org/bikeshed/bikeshed.py", line 6, in 
    cli.main()
  File "/sites/api.csswg.org/bikeshed/bikeshed/cli.py", line 466, in main
    handleSpec(options, extras)
  File "/sites/api.csswg.org/bikeshed/bikeshed/cli.py", line 515, in handleSpec
    lineNumbers=options.lineNumbers,
  File "/sites/api.csswg.org/bikeshed/bikeshed/Spec.py", line 61, in __init__
    self.inputSource = InputSource(inputFilename, chroot=constants.chroot)
TypeError: __init__() got an unexpected keyword argument 'chroot'
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20WICG/urlpattern%2366.)._
</details>

PR #67, opened 2021-07-30
Spec the constructor except for the constructor string parser.

wanderview:
@domenic PTAL.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/67.html" title="Last updated on Aug 11, 2021, 9:05 PM UTC (a685665)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/67/83a6925...a685665.html" title="Last updated on Aug 11, 2021, 9:05 PM UTC (a685665)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -29,6 +29,7 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
   type: abstract-op
     text: Get; url: #sec-get-o-p
     text: RegExpBuiltinExec; url: #sec-regexpbuiltinexec
+    text: RegExpCreate; url: #sec-regexpcreate
     text: ToString; url: #sec-tostring
   type: dfn
     text: IdentifierPart; url: #prod-IdentifierPart
@@ -38,11 +39,13 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
 spec: url; urlPrefix: https://url.spec.whatwg.org/
   type: dfn
     text: cannot be a base URL path state; url: #cannot-be-a-base-url-path-state
+    text: default port; url: #default-port
     text: fragment state; url: #fragment-state
     text: hostname state; url: #hostname-state
     text: path start state; url: #path-start-state
     text: port state; url: #port-state
     text: query state; url: #query-state
+    text: special scheme; url: #special-scheme
     text: scheme start state; url: #scheme-start-state
 </pre>
 
@@ -165,21 +168,21 @@ dictionary URLPatternComponentResult {
 };
 </xmp>
 
-Each {{URLPattern}} object has an associated <dfn for=URLPattern>protocol component</dfn>, a [=component=], initially null.
+Each {{URLPattern}} object has an associated <dfn for=URLPattern>protocol component</dfn>, a [=component=], which must be set upon creation.
 
-Each {{URLPattern}} object has an associated <dfn for=URLPattern>username component</dfn>, a [=component=], initially null.
+Each {{URLPattern}} object has an associated <dfn for=URLPattern>username component</dfn>, a [=component=], which must be set upon creation.
 
-Each {{URLPattern}} object has an associated <dfn for=URLPattern>password component</dfn>, a [=component=], initially null.
+Each {{URLPattern}} object has an associated <dfn for=URLPattern>password component</dfn>, a [=component=], which must be set upon creation.
 
-Each {{URLPattern}} object has an associated <dfn for=URLPattern>hostname component</dfn>, a [=component=], initially null.
+Each {{URLPattern}} object has an associated <dfn for=URLPattern>hostname component</dfn>, a [=component=], which must be set upon creation.
 
-Each {{URLPattern}} object has an associated <dfn for=URLPattern>port component</dfn>, a [=component=], initially null.
+Each {{URLPattern}} object has an associated <dfn for=URLPattern>port component</dfn>, a [=component=], which must be set upon creation.
 
-Each {{URLPattern}} object has an associated <dfn for=URLPattern>pathname component</dfn>, a [=component=], initially null.
+Each {{URLPattern}} object has an associated <dfn for=URLPattern>pathname component</dfn>, a [=component=], which must be set upon creation.
 
-Each {{URLPattern}} object has an associated <dfn for=URLPattern>search component</dfn>, a [=component=], initially null.
+Each {{URLPattern}} object has an associated <dfn for=URLPattern>search component</dfn>, a [=component=], which must be set upon creation.
 
-Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component</dfn>, a [=component=], initially null.
+Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component</dfn>, a [=component=], which must be set upon creation.
 
 <dl class="domintro non-normative">
   <dt><code>{{URLPattern}} . {{URLPattern/protocol}}</code>
@@ -222,6 +225,30 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
     <p>The normalized hash pattern string.
   </dd>
 </dl>
+
+<div algorithm>
+  The <dfn constructor for=URLPattern lt="URLPattern(input, baseURL)">new URLPattern(|input|, |baseURL|)</dfn> constructor steps are:
+
+  1. Let |init| be null.
+  1. If |input| is a [=scalar value string=] then:
+    1. Set |init| to the result of running [=parse a constructor string=] given |input|.
+    1. Set |init|["{{URLPatternInit/baseURL}}"] to |baseURL|.
+  1. Else:
+    1. [=Assert=]: |input| is a {{URLPatternInit}}.
+    1. If |baseURL| is given, then throw a {{TypeError}}.
+    1. Set |init| to |input|.
+  1. Let |processedInit| be the result of [=process a URLPatternInit=] given |init|, "`pattern`", null, null, null, null, null, null, null, and null.
+  1. If |processedInit|["{{URLPatternInit/protocol}}"] is a [=special scheme=] and |processedInit|["{{URLPatternInit/port}}"] is its corresponding [=default port=], then set |processedInit|["{{URLPatternInit/port}}"] to the empty string.
+  1. Set [=this=]'s [=URLPattern/protocol component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/protocol}}"], [=canonicalize a protocol=], and [=default options=].
+  1. Set [=this=]'s [=URLPattern/username component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/username}}"], [=canonicalize a username=], and [=default options=].
+  1. Set [=this=]'s [=URLPattern/password component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/password}}"], [=canonicalize a password=], and [=default options=].
+  1. Set [=this=]'s [=URLPattern/hostname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/hostname}}"], [=canonicalize a hostname=], and [=hostname options=].
+  1. Set [=this=]'s [=URLPattern/port component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/port}}"], [=canonicalize a port=], and [=default options=].
+  1. If the result of running [=protocol component matches a special scheme=] given [=this=]'s [=URLPattern/protocol component=] is true, then set [=this=]'s [=URLPattern/pathname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/pathname}}"], [=canonicalize a standard pathname=], and [=standard pathname options=].
+  1. Else set [=this=]'s [=URLPattern/pathname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/pathname}}"], [=canonicalize a cannot-be-a-base-URL pathname=], and [=default options=]
+  1. Set [=this=]'s [=URLPattern/search component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/search}}"], [=canonicalize a search=], and [=default options=].
+  1. Set [=this=]'s [=URLPattern/hash component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/hash}}"], [=canonicalize a hash=], and [=default options=].
+</div>
 
 <div algorithm>
   The <dfn attribute for="URLPattern">protocol</dfn> getter steps are:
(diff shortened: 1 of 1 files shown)


wanderview:
This depends on #66.

spec.bs
```diff
@@ -223,6 +227,48 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Set |protocol| to |processedInit|["{{URLPatternInit/protocol}}"].
+  1. Set |username| to |processedInit|["{{URLPatternInit/username}}"].
+  1. Set |password| to |processedInit|["{{URLPatternInit/password}}"].
+  1. Set |hostname| to |processedInit|["{{URLPatternInit/hostname}}"].
+  1. Set |port| to |processedInit|["{{URLPatternInit/port}}"].
+  1. Set |pathname| to |processedInit|["{{URLPatternInit/pathname}}"].
+  1. Set |search| to |processedInit|["{{URLPatternInit/search}}"].
+  1. Set |hash| to |processedInit|["{{URLPatternInit/hash}}"].
+  1. If |protocol| is a [=special scheme=] and |port| is its corresponding [=url/port=], then set |port| to the empty string.
+  1. Let |protocol component| be the result of running [=compile a component=] given |protocol|, [=canonicalize a protocol=], and [=default options=].
```
domenic, line 259: FYI you could do `be the result of [=compiling a component=]` etc.

spec.bs
```diff
@@ -223,6 +227,48 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Let |username component| be the result of running [=compile a component=] given |username|, [=canonicalize a username=], and [=default options=].
+  1. Let |password component| be the result of running [=compile a component=] given |password|, [=canonicalize a password=], and [=default options=].
+  1. Let |hostname component| be the result of running [=compile a component=] given |hostname|, [=canonicalize a hostname=], and [=hostname options=].
+  1. Let |port component| be the result of running [=compile a component=] given |port|, [=canonicalize a port=], and [=default options=].
+  1. Let |pathname component| be null.
+  1. If the result of running [=protocol component matches a special scheme=] given |protocol component|, then set |pathname component| to the result of running [=compile a component=] given |pathname|, [=canonicalize a standard pathname=], and [=standard pathname options=].
```
domenic, line 265:
```suggestion
  1. If the result of running [=protocol component matches a special scheme=] given |protocol component| is true, then set |pathname component| to the result of running [=compile a component=] given |pathname|, [=canonicalize a standard pathname=], and [=standard pathname options=].
```

spec.bs
```diff
@@ -223,6 +227,48 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Else set |pathname component| to the result of running [=compile a component=] given |pathname, [=canonicalize a cannot-be-a-base-URL pathname=], and [=default options=]
+  1. Let |search component| be the result of running [=compile a component=] given |search|, [=canonicalize a search=], and [=default options=].
+  1. Let |hash component| be the result of running [=compile a component=] given |hash|, [=canonicalize a hash=], and [=default options=].
+  1. Return a new {{URLPattern}} whose [=URLPattern/protocol component=] is |protocol component|, [=URLPattern/username component=] is |username component|, [=URLPattern/password component=] is |password component|, [=URLPattern/hostname component=] is |hostname component|, [=URLPattern/port component=] is |port component|, [=URLPattern/pathname component=] is |pathname component|, [=URLPattern/search component=] is |search component|, and [=URLPattern/hash component=] is |hash component|.
```
domenic, line 269: As of about a year ago constructors no longer return new objects; instead they mutate `[=this=]`. So you want something like `Set [=this=]'s [=URLPattern/protocol component=] to ...`

You could probably remove the intermediate variables _protocol component_ etc. by just setting the appropriate field of `[=this=]` in steps 259 onward.

spec.bs
```diff
@@ -289,11 +335,22 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
 A {{URLPattern}} is associated with multiple <dfn>component</dfn> [=structs=].
 
-A [=component=] has an associated <dfn for=component>pattern string</dfn>, a [=pattern string/well formed=] [=/pattern string=], initially null.
+A [=component=] has an associated <dfn for=component>pattern string</dfn>, a [=pattern string/well formed=] [=/pattern string=], which must be set upon creation.
 
-A [=component=] has an associated <dfn for=component>regular expression</dfn>, a {{RegExp}}, initially null.
+A [=component=] has an associated <dfn for=component>regular expression</dfn>, a {{RegExp}}, which must be set upon creation.
 
-A [=component=] has an associated <dfn for=component>group name list</dfn>, an array of strings, initially an empty array.
+A [=component=] has an associated <dfn for=component>group name list</dfn>, an array of strings, which must be set upon creation.
```
domenic, line 342: While here:
```suggestion
A [=component=] has an associated <dfn for=component>group name list</dfn>, a [=list=] of strings, which must be set upon creation.
```

spec.bs
```diff
@@ -289,11 +335,22 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+
+<div algorithm>
+  To <dfn>compile a component</dfn> given a string |input|, [=/options=] |options|, and [=/encoding callback=] |encoding callback|:
+
+  1. Let |part list| be the result of running [=parse a pattern string=] given |input|, |options|, and |encoding callback|.
+  1. Let (|regular expression string|, |name list|) be the result of running [=generate a regular expression and name list=] given |part list| and |options|.
+  1. Let |regular expression| be [$RegExpCreate$](|regular expression string|, "`u`").
+  1. If |regular expression| is an [=abrupt completion=], throw a {{TypeError}}.
```
domenic, line 350: I suggest instead adding a line to the previous bullet which says "If this throws an exception, catch it, and throw a TypeError".

(Or, it might be nicer to just let it throw the original SyntaxError? That seems a bit more correct anyway.)

Reasoning: when dealing with ES abstract operations, you can either do what the ES spec does, and treat them as always returning completion values, or you can treat them like web platform algorithms, which throw exceptions/return values. If you do the ES spec way (which is done by Streams, Web IDL, and some sections of HTML) then you need to do some more hoops, e.g. to extract the actual return value you'd need to get `|regular expression|.[[Value]]`, and you'd need to update other parts of your spec to use the ! or ? prefixes.

spec.bs
```diff
@@ -406,6 +463,30 @@ A [=component=] has an associated <dfn for=component>group name list</dfn>, an a
...
+The <dfn>hostname options</dfn> is an [=options=] [=struct=] with [=options/delimiter code point=] set "`.`" and [=options/prefix code point=] set to the empty string.
+
+The <dfn>standard pathname options</dfn> is an [=options=] [=struct=] with [=options/delimiter code point=] set "`/`" and [=options/prefix code point=] set to "`/`".
+
+<div algorithm>
+  To determine if a <dfn>protocol component matches a special scheme</dfn> given a [=component=] |protocol component|:
+
+  1. Let |special scheme list| be a [=list=] populated with all of the [=special scheme=] values.
+  1. [=list/For each=] |scheme| of |special scheme list|:
+    1. Let |test result| be [$RegExpBuiltinTest$](|protocol component|'s [=component/regular expression=], |scheme|).
```
domenic, line 477: RegExpBuiltinTest doesn't seem to exist. I think you want to use BuiltinExec and check for null.

spec.bs
```diff
@@ -1171,11 +1252,7 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   To <dfn>process pathname for init</dfn> given a string |pathnameValue|, a string |protocolValue|, and a string |type|:
 
   1. If |type| is "`pattern`" then return |pathnameValue|.
-  1. Let |dummyURL| be a new [=URL record=].
-  1. If |protocolValue| is null or the empty string, then set |protocolValue| to "`https`".
-  1. Set |dummyURL|'s [=url/scheme=] to |protocolValue|.
-     <p class="note">Note, we set the [=URL record=]'s [=url/scheme=] in order to check if it [=is special=].  If |protocolValue| is missing we purposely default the URL to being special by using "`https`" for the [=url/scheme=].</p>
-  1. If |dummyURL| [=is special=], then:
+  1. If |protocolValue| is a [=special scheme=], then:
```
domenic, line 1255: Should this be part of the previous PR?

spec.bs
```diff
@@ -223,6 +227,48 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+
+  1. Let |init| be null.
+  1. If |input| is a string then:
+    1. Set |init| to the result of running [=parse a constructor string=] given |input|.
+    1. Set |init|["{{URLPatternInit/baseURL}}"] to |baseURL|.
+  1. Else:
+    1. [=Assert=]: |input| is a {{URLPatternInit}}.
+    1. If |baseURL| is given, then throw a {{TypeError}}.
+    1. Set |init| to |input|.
+  1. Let |protocol| be null.
```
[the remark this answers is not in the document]
wanderview: But the passed in values are significant; empty string vs null.  Also, this follows the pattern already in the match algorithm.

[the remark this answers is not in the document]
domenic: Well, right now it's never null, because you always overwrite it on line 250.

spec.bs
wanderview: Allowing the original error out will probably give an error with a hard to understand message string.  The caller won't be able to see the compiled regexp string, so I don't know that it is useful.

Also, the implementation today throws a TypeError like we do for other invalid patterns.  Are you suggesting those should all be syntax errors?

spec.bs
wanderview: I implemented your suggestion, but do I need to change the part that says things like [=URLPattern/protocol component=] "must be set upon creation"?  Its a bit weird I'm setting them in this algorithm when they have an undefined value.

spec.bs
wanderview: I could try to move it back, but I did it here since I figured out the better way to do the "is special" check.  If its ok, can I just leave it here instead of trying to move between the dependent branches?

spec.bs
domenic: I think that is still good; the phrase isn't super well-defined but the intention is that whenever a new URLPattern is created, those things need to be set before any JS code could read them (or cause the spec to read them).

spec.bs
domenic: Yeah, fair point about the original message.

SyntaxError vs. TypeError is an interesting (although not super-important) question. TypeError seems good for now, but I've asked in the TC39 matrix channel what they think. It seems like sometimes the ES spec does use SyntaxError for generic "parsing errors", e.g. in JSON.parse or when you do `BigInt("NaN")`, as well as in RegExp parsing.

spec.bs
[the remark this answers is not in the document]
wanderview: These values are passed as arguments to [=process a URLPatternInit=].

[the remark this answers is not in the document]
domenic: Oh yeah... that makes no sense to me then, since you're always passing null. That algorithm doesn't even accept null; it requires those arguments to be strings.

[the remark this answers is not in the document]
wanderview: I think that algorithm has been changed to accept string or null now.  As we talked about offline, though, I will plan to do another pass to try to make "process a URLPatternInit" cleaner.  Its a bit awkward at the moment due to the translation from c++ where these arguments are passed by reference and modified in place.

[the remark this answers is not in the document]
domenic: Even before that refactoring, I think this would be clearer if you got rid of lines 240-247, changed line 248 to just say "null, null, null, null, null, null, null, and null", and then used "Let .. be" instead of "Set ... to" on lines 249-256.

It's just really confusing right now to reuse the variables as both inputs and outputs.

spec.bs
```diff
@@ -223,6 +226,46 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Let |processedInit| be the result of [=process a URLPatternInit=] given |init|, "`pattern`", |protocol|, |username|, |password|, |hostname|, |port|, |pathname|, |search|, and |hash|.
+  1. Set |protocol| to |processedInit|["{{URLPatternInit/protocol}}"].
+  1. Set |username| to |processedInit|["{{URLPatternInit/username}}"].
+  1. Set |password| to |processedInit|["{{URLPatternInit/password}}"].
+  1. Set |hostname| to |processedInit|["{{URLPatternInit/hostname}}"].
+  1. Set |port| to |processedInit|["{{URLPatternInit/port}}"].
+  1. Set |pathname| to |processedInit|["{{URLPatternInit/pathname}}"].
+  1. Set |search| to |processedInit|["{{URLPatternInit/search}}"].
+  1. Set |hash| to |processedInit|["{{URLPatternInit/hash}}"].
+  1. If |protocol| is a [=special scheme=] and |port| is its corresponding [=url/port=], then set |port| to the empty string.
```
domenic, line 257: It looks like there's a definition for this: https://url.spec.whatwg.org/#default-port

spec.bs
```diff
@@ -372,6 +426,30 @@ A [=component=] has an associated <dfn for=component>group name list</dfn>, a [=
...
+The <dfn>default options</dfn> is an [=options=] [=struct=] with [=options/delimiter code point=] set to the empty string and [=options/prefix code point=] set to the empty string.
+
+The <dfn>hostname options</dfn> is an [=options=] [=struct=] with [=options/delimiter code point=] set "`.`" and [=options/prefix code point=] set to the empty string.
+
+The <dfn>standard pathname options</dfn> is an [=options=] [=struct=] with [=options/delimiter code point=] set "`/`" and [=options/prefix code point=] set to "`/`".
+
+<div algorithm>
+  To determine if a <dfn>protocol component matches a special scheme</dfn> given a [=component=] |protocol component|:
+
+  1. Let |special scheme list| be a [=list=] populated with all of the [=special scheme=] values.
```
domenic, line 438: "special scheme values" is a bit unclear. Maybe just "all the special schemes"?

spec.bs
```diff
@@ -223,6 +226,46 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
+  1. Set [=this=]'s [=URLPattern/username component=] to the result of [=compiling a component=] given |username|, [=canonicalize a username=], and [=default options=].
+  1. Set [=this=]'s [=URLPattern/password component=] to the result of [=compiling a component=] given |password|, [=canonicalize a password=], and [=default options=].
+  1. Set [=this=]'s [=URLPattern/hostname component=] to the result of [=compiling a component=] given |hostname|, [=canonicalize a hostname=], and [=hostname options=].
+  1. Set [=this=]'s [=URLPattern/port component=] to the result of [=compiling a component=] given |port|, [=canonicalize a port=], and [=default options=].
+  1. If the result of running [=protocol component matches a special scheme=] given [=this=]'s [=URLPattern/protocol component=] is true, then set [=this=]'s [=URLPattern/pathname component=] to the result of [=compiling a component=] given |pathname|, [=canonicalize a standard pathname=], and [=standard pathname options=].
+  1. Else set [=this=]'s [=URLPattern/pathname component=] to the result of [=compiling a component=] given |pathname, [=canonicalize a cannot-be-a-base-URL pathname=], and [=default options=]
```
domenic, line 264:
```suggestion
  1. Else set [=this=]'s [=URLPattern/pathname component=] to the result of [=compiling a component=] given |pathname|, [=canonicalize a cannot-be-a-base-URL pathname=], and [=default options=]
```

spec.bs
domenic: Regardless of which way you go here, you should fix the editorial issue here with abrupt completion language vs. exception language.

TC39 people said this seems like a good use case for `{{SyntaxError}}` instead of `{{TypeError}}`, so that might be a good thing to change, either here or in a followup coordinated with the implementation/tests.

spec.bs
```diff
@@ -1175,7 +1253,7 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   1. If |protocolValue| is null or the empty string, then set |protocolValue| to "`https`".
   1. Set |dummyURL|'s [=url/scheme=] to |protocolValue|.
      <p class="note">Note, we set the [=URL record=]'s [=url/scheme=] in order to check if it [=is special=].  If |protocolValue| is missing we purposely default the URL to being special by using "`https`" for the [=url/scheme=].</p>
-  1. If |dummyURL| [=is special=], then return the result of running [=canonicalize a standard pathname=] given |pathnameValue|.
+  1. If |protocolValue| is a [=special scheme=], then return the result of running [=canonicalize a standard pathname=] given |pathnameValue|.
```
domenic, line 1256: Hmm, can we just get rid of all steps related to dummyURL? It looks like you're just creating it but never using it.

spec.bs
wanderview: Tried to fix.  Filed #74 to follow-up on exception type.

wanderview:
Updated.  PTAL.

domenic (approved):
Very nice

wanderview:
Ugh, this merged into dev-encoding instead of main.

PR #69, opened 2021-07-30
Spec the constructor string parser.

wanderview:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/69.html" title="Last updated on Aug 12, 2021, 3:13 PM UTC (493de8e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/69/26710dc...493de8e.html" title="Last updated on Aug 12, 2021, 3:13 PM UTC (493de8e)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -427,10 +427,321 @@ The <dfn>standard pathname options</dfn> is an [=options=] [=struct=] with [=opt
 
 <h3 id=constructor-string-parsing>Constructor String Parsing</h3>
 
+A <dfn export>constructor string parser</dfn> is a [=struct=].
+
+A [=constructor string parser=] has an associated <dfn export for="constructor string parser">input</dfn>, a string, which must be set upon creation.
+
+A [=constructor string parser=] has an associated <dfn export for="constructor string parser">token list</dfn>, a [=/token list=], which must be set upon creation.
+
+A [=constructor string parser=] has an associated <dfn export for="constructor string parser">result</dfn>, a {{URLPatternInit}}, initially set to a new {{URLPatternInit}}.
+
+A [=constructor string parser=] has an associated <dfn export for="constructor string parser">component start</dfn>, a number, initially set to 0.
+
+A [=constructor string parser=] has an associated <dfn export for="constructor string parser">token index</dfn>, a number, initially set to 0.
+
+A [=constructor string parser=] has an associated <dfn export for="constructor string parser">token increment</dfn>, a number, initially set to 1.
+
+A [=constructor string parser=] has an associated <dfn export for="constructor string parser">group depth</dfn>, a number, initially set to 0.
+
+A [=constructor string parser=] has an associated <dfn export for="constructor string parser">should treat as a standard URL</dfn>, a boolean, initially set to false.
+
+A [=constructor string parser=] has an associated <dfn export for="constructor string parser">state</dfn>, a string, initially set to "<a for="constructor string parser/state">`init`</a>".  It must be one of the following:
+
+<ul>
+  <li>"<dfn for="constructor string parser/state">`init`</dfn>"</li>
+  <li>"<dfn export for="constructor string parser/state">`protocol`</dfn>"</li>
+  <li>"<dfn export for="constructor string parser/state">`authority`</dfn>"</li>
+  <li>"<dfn export for="constructor string parser/state">`username`</dfn>"</li>
+  <li>"<dfn export for="constructor string parser/state">`password`</dfn>"</li>
+  <li>"<dfn export for="constructor string parser/state">`hostname`</dfn>"</li>
+  <li>"<dfn export for="constructor string parser/state">`port`</dfn>"</li>
+  <li>"<dfn export for="constructor string parser/state">`pathname`</dfn>"</li>
+  <li>"<dfn export for="constructor string parser/state">`search`</dfn>"</li>
+  <li>"<dfn export for="constructor string parser/state">`hash`</dfn>"</li>
+  <li>"<dfn export for="constructor string parser/state">`done`</dfn>"</li>
+</ul>
+
+<div class=note>
+  <p>The URLPattern constructor string algorithm is very similar to the [=basic URL parser=] algorithm, but some differences prevent us from using that algorithm directly.
+  <p>First, the URLPattern constructor string parser operates on [=tokens=] generated using the "`lenient`" [=tokenize policy=].  In constrast, [=basic URL parser=] operates code points.  Operating on [=tokens=] allows the URLPattern constructor string parser to more easily distinguish between code points that are significant pattern syntax and code points that might be a URL component separator.  For example, it makes it trivial to handle named groups like "`:hmm`" in "`https://a.c:hmm.example.com:8080`" without getting confused with the port number.
+  <p>Second, the URLPattern constructor string parser needs to avoid applying URL canonicalization to all code points like [=basic URL parser=] does.  Instead we perform canonicalization on only parts of the pattern string we know are safe later when compiling each component pattern string.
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic PTAL, although it can wait until after the other PRs are merged.  I think this is the last substantive PR.  After this I think it will mostly be polish, non-normative prose, etc.

spec.bs
```diff
@@ -427,10 +427,325 @@ The <dfn>standard pathname options</dfn> is an [=options=] [=struct=] with [=opt
...
+  <p>Second, the URLPattern constructor string parser needs to avoid applying URL canonicalization to all code points like [=basic URL parser=] does.  Instead we perform canonicalization on only parts of the pattern string we know are safe later when compiling each component pattern string.
+  <p class=allow-2119>Finally, the URLPattern constructor string parser does not handle some parts of the [=basic URL parser=] state machine.  For example, it does not treat back slashes specially as they would all be treated as pattern characters and would require excessive escaping.  In addition, this parser may not handle some more esoteric parts of the URL parsing algorithm like file URLs with a hostname.  The goal with this parser was to handle the most common URLs while allowing any niche case to be handled instead via the {{URLPatternInit}} constructor.
```
domenic, line 468:
```suggestion
  <p class=allow-2119>Finally, the URLPattern constructor string parser does not handle some parts of the [=basic URL parser=] state machine.  For example, it does not treat backslashes specially as they would all be treated as pattern characters and would require excessive escaping.  In addition, this parser may not handle some more esoteric parts of the URL parsing algorithm like file URLs with a hostname.  The goal with this parser was to handle the most common URLs while allowing any niche case to be handled instead via the {{URLPatternInit}} constructor.
```

spec.bs
```diff
@@ -427,10 +427,325 @@ The <dfn>standard pathname options</dfn> is an [=options=] [=struct=] with [=opt
...
+      <p>To implement this we initialize components in |parser|'s [=constructor string parser/result=] with empty string in advance.
+      <p>We can't, however, do this immediately.  We want to allow the `baseURL` to provide information for relative URLs, so we only want to set the default empty string values for components following the first component in the relative URL.  We therefore wait to set the default component values until after we exit the "<a for="constructor string parser/state">`init`</a>" [=constructor string parser/state=].
+    </div>
+  1. [=While=] |parser|'s [=constructor string parser/token index=] is less than |parser|'s [=constructor string parser/token list=] [=list/size=]:
+    1. Set |parser|'s [=constructor string parser/token increment=] to 1.
+      <p class="note allow-2119">On every iteration of the parse loop the |parser|'s [=constructor string parser/token index=] should be incremented by its [=constructor string parser/token increment=] value.  Typically this means incrementing by 1, but at certain times it is set to zero.  The [=constructor string parser/token increment=] is then always reset back to 1 at the top of the loop.
```
domenic, line 482:
```suggestion
      <p class="note allow-2119">On every iteration of the parse loop the |parser|'s [=constructor string parser/token index=] will be incremented by its [=constructor string parser/token increment=] value.  Typically this means incrementing by 1, but at certain times it is set to zero.  The [=constructor string parser/token increment=] is then always reset back to 1 at the top of the loop.
```

spec.bs
```diff
@@ -427,10 +427,325 @@ The <dfn>standard pathname options</dfn> is an [=options=] [=struct=] with [=opt
...
+      1. If |parser|'s [=constructor string parser/state=] is "<a for="constructor string parser/state">`init`</a>":
+        <div class=note>
+          <p class=allow-2119>If we reached the end of the string in the "<a for="constructor string parser/state">`init`</a>" [=constructor string parser/state=], then we failed to find a protocol terminator and this must be a relative URLPattern constructor string.
+        </div>
+        1. Run [=rewind=] given |parser|.
+          <p class=note>We next determine at which component the relative pattern begins.  Relative pathnames are most common, but URLs and URLPattern constructor strings can begin with the search or hash components as well.
+        1. If the result of running [=is a hash prefix=] given |parser| is true, then run [=change state=] given |parser|, "<a for="constructor string parser/state">`hash`</a>" and 1.
+        1. Else if the result of running [=is a search prefix=] given |parser| is true:
+          1. Run [=change state=] given |parser|, "<a for="constructor string parser/state">`search`</a>" and 1.
+          1. Set |parser|'s [=constructor string parser/result=]["`hash`"] to the empty string.
```
domenic, line 493:
```suggestion
          1. Set |parser|'s [=constructor string parser/result=]["{{URLPatternInit/hash}}"] to the empty string.
```

same for several other cases.

spec.bs
```diff
@@ -427,10 +427,325 @@ The <dfn>standard pathname options</dfn> is an [=options=] [=struct=] with [=opt
...
+        </dd>
+        <dt>"<a for="constructor string parser/state">`authority`</a>"</dt>
+        <dd>
+          1. If the result of running [=is an identity terminator=] given |parser| is true, then run [=rewind and set state=] given |parser| and "<a for="constructor string parser/state">`username`</a>".
+          1. Else if any of the following are true, then run [=rewind and set state=] given |parser| and "<a for="constructor string parser/state">`hostname`</a>".
+            <ul>
+              <li>The result of running [=is a pathname start=] given |parser|.
+              <li>The result of running [=is a search prefix=] given |parser|.
+              <li>The result of running [=is a hash prefix=] given |parser|.
+            </ul>
```
domenic, line 557: Nit: I personally like splitting these up, i.e.

> Else if any of the following are true:
> 
> * lowercase sentence fragment;
> * lowercase sentence fragment; or
> * lowercase sentence fragment,
> 
> then run...

spec.bs
```diff
@@ -427,10 +427,325 @@ The <dfn>standard pathname options</dfn> is an [=options=] [=struct=] with [=opt
...
+
+<div algorithm>
+To <dfn>change state</dfn> given a [=constructor string parser=] |parser|, a [=constructor string parser/state=] |state|, and a number |skip|:
+
+  1. If |state| is not "<a for="constructor string parser/state">`init`</a>", not "<a for="constructor string parser/state">`authority`</a>", and not "<a for="constructor string parser/state">`done`</a>", then set |parser|'s [=constructor string parser/result=][|state|] to the result of running [=make a component string=] given |parser|.
+  1. Run [=change state without setting the component=] given |parser|, |state|, and |skip|.
+</div>
+
+<div algorithm>
+To <dfn>change state without setting the component</dfn> give a [=constructor string parser=] |parser|, a [=constructor string parser/state=] |state|, and a number |skip|:
```
domenic, line 611: This only has the previous algorithm as a caller, so probably is best inlined into it?

spec.bs
```diff
@@ -427,10 +427,325 @@ The <dfn>standard pathname options</dfn> is an [=options=] [=struct=] with [=opt
...
+  1. Return |token|.
+</div>
+
+<div algorithm>
+To run <dfn>is a non-special pattern char</dfn> given a [=constructor string parser=] |parser|, a number |index|, and a string |value|:
+
+  1. Let |token| be the result of running [=get a safe token=] given |parser| and |index|.
+  1. If |token|'s [=token/value=] is not |value|, then return false.
+  1. If any of the following are true, then return true:
+    <ol>
```
domenic, line 650: See my above nit about style, but even if you don't go that route, this should be a ul, not ol.

spec.bs
```diff
@@ -498,7 +813,7 @@ A [=tokenizer=] has an associated <dfn for=tokenizer>code point</dfn>, a Unicode
   1. Let |tokenizer| be a new [=tokenizer=].
   1. Set |tokenizer|'s [=tokenizer/input=] to |input|.
   1. Set |tokenizer|'s [=tokenizer/policy=] to |policy|.
-  1. While |tokenizer|'s [=tokenizer/index=] is less than |tokenizer|'s [=tokenizer/input=]'s [=string/length=]:
+  1. While |tokenizer|'s [=tokenizer/index=] is less than |tokenizer|'s [=tokenizer/input=]'s [=string/code point length=]:
```
domenic, line 816: Good catch!

domenic (approved):
Although I can't claim to have run the algorithms in my head and verified they're correct, overall it looks pretty solid! Only a few editorial suggestions and nits.

spec.bs
wanderview: Huh, its not called separately in the c++ either.  It definitely was before.  I must have gotten rid of that callsite in a refactor at some point.  I'll inline here.

PR #85, opened 2021-08-12
Add a note that implementations do not have to use regexps to match.

wanderview:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/85.html" title="Last updated on Aug 12, 2021, 8:00 PM UTC (eaf49cc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/85/68b66a0...eaf49cc.html" title="Last updated on Aug 12, 2021, 8:00 PM UTC (eaf49cc)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -328,7 +328,7 @@ A [=component=] has an associated <dfn for=component>group name list</dfn>, a [=
   1. Let |part list| be the result of running [=parse a pattern string=] given |input|, |options|, and |encoding callback|.
   1. Let (|regular expression string|, |name list|) be the result of running [=generate a regular expression and name list=] given |part list| and |options|.
   1. Let |regular expression| be [$RegExpCreate$](|regular expression string|, "`u`").  If this throws an exception, catch it, and throw a {{TypeError}}.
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic a small PR adding a note about match algorithm.  FYI in case you want to review.

spec.bs
```diff
@@ -328,6 +328,7 @@ A [=component=] has an associated <dfn for=component>group name list</dfn>, a [=
   1. Let |part list| be the result of running [=parse a pattern string=] given |input|, |options|, and |encoding callback|.
   1. Let (|regular expression string|, |name list|) be the result of running [=generate a regular expression and name list=] given |part list| and |options|.
   1. Let |regular expression| be [$RegExpCreate$](|regular expression string|, "`u`").  If this throws an exception, catch it, and throw a {{TypeError}}.
+    <p class="note allow-2119">The specification uses regular expressions to perform all matching, but this is not required.  Implementations are free to perform matching directly against the [=/part list=] when possible; e.g. when there are no custom regexp matching groups.  If there are custom regular expressions, however, its important that they should be immediately evaluated in [=compile a component=] algorithm so an error can be thrown if they are invalid.
```
domenic, line 331:
```suggestion
    <p class="note allow-2119">The specification uses regular expressions to perform all matching, but this is not required.  Implementations are free to perform matching directly against the [=/part list=] when possible; e.g. when there are no custom regexp matching groups.  If there are custom regular expressions, however, its important that they should be immediately evaluated in the [=compile a component=] algorithm so an error can be thrown if they are invalid.
```

domenic (approved):

Issue #35, opened 2020-12-15
passing base URL as a dictionary property vs a separate string argument


wanderview:
So far I've been mostly aiming to have the base URL, if present, passed as a property on the dictionary argument.  For example:

```javascript
new URLPattern({ pathname: '/foo/bar', baseURL: 'https://example.com' });
```

And:

```javascript
p.test({ pathname: '/foo/bar', baseURL: 'https://example.com' });
```

I also have a TODO, though, to support a baseURL for the string URL input to test()/exec().  For example:

```javascript
p.test('/foo/bar', 'https://example.com');
```

@blakeembrey suggested maybe we should always support base URL as a second argument.  For example, the first two examples would look like this:

```javascript
new URLPattern({ pathname: '/foo/bar' }, 'https://example.com');

p.test({ pathname: '/foo/bar' }, 'https://example.com');
```

This also seems reasonable to me.  I think the one downside is that you lose the "named arguments" style of the dictionary.  We could possibly support both styles, though, and let developers choose which they prefer.

Thoughts?

jeremyroman:
FWIW I don't have a strong preference but I think allowing it to existing in both places is potentially confusing. What does `new URLPattern({ pathname: '/foo/bar', baseURL: 'https://example.com/'}, 'https://example.net/')` do?

wanderview:
I don't support that for constructor currently.  But I did just add support for it to test() and exec().  The webidl looks like:

```
boolean test(URLPatternInit input);
boolean test(USVString input, optional USVString baseURL);
```

Which means passing a second argument will coerce the first argument to a string.  So this will always return false:

```javascript
pattern.test({ pathname: '/foo/bar' }, 'https://example.com');
```

I could at least get this to an error message if I changed the webidl to:

```
typedef (USVString or URLPatternInit) URLPatternInput;
boolean test(URLPatternInput input, optional USVString baseURL);
```

But I chose not to do that since it seemed better to defer to webidl semantics.  I don't feel too strongly here if other think we should go a different way.

wanderview:
Talked with @domenic this morning and he suggested we go with the union type to provide a more meaningful exception.  It provides a better experience for developers at the cost of a slightly harder to read spec.  We can add a note to the spec, though.

I think we should not support second baseURL argument for structured input for now.  We will throw an exception if the user tries.  If we get lots of requests for the feature, though, we can re-assess.

wanderview:
This is now codified in the spec.

PR #86, opened 2021-08-12
Update the domintro section. (Fixes #76)

### LINKED ISSUE #76: Finish and tidy up the domintro section
The domintro section currently lists the properties only; it doesn't list the constructor or methods.

Also, the way the properties are listed is a bit strange; it makes them look like static properties: `URLPattern.protocol`. Instead they should look like instance properties: <code><var>urlPattern</var>.protocol</code>.

See https://github.com/whatwg/meta/issues/190, or probably just skip to https://github.com/whatwg/meta/issues/190#issuecomment-799601600 for my summary of best practices that we're trying to converge the ecosystem on a bit.

wanderview:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/86.html" title="Last updated on Aug 12, 2021, 11:34 PM UTC (ff3d478)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/86/756df36...ff3d478.html" title="Last updated on Aug 12, 2021, 11:34 PM UTC (ff3d478)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -188,32 +188,38 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
 <dl class="domintro non-normative">
   <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL)|URLPattern}}(|input|)</code></dt>
   <dd>
-    Constructs a new URLPattern object.  The |input| is an object containing separate patterns for each URL component; e.g. hostname, pathname, etc.  Missing components will default to a wildcard pattern.  In addition, |input| can contain a baseURL property that provides static text patterns for any missing components.
+    Constructs a new {{URLPattern}} object.  The |input| is an object containing separate patterns for each URL component; e.g. hostname, pathname, etc.  Missing components will default to a wildcard pattern.  In addition, |input| can contain a {{URLPatternInit/baseURL}} property that provides static text patterns for any missing components.
   </dd>
 
-  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL)|URLPattern}}(|input|, |baseURL|)</code></dt>
+  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL)|URLPattern}}(|patternString|, |baseURL|)</code></dt>
   <dd>
-    Constructs a new URLPattern object.  The |input| is a URL string containing pattern syntax for one or more components.  If |baseURL| is provided, then |input| can be relative.  This constructor will always set at least an empty string value and does not default any components to wildcard patterns.
+    Constructs a new {{URLPattern}} object.  |patternString| is a URL string containing pattern syntax for one or more components.  If |baseURL| is provided, then |patternString| can be relative.  This constructor will always set at least an empty string value and does not default any components to wildcard patterns.
   </dd>
 
   <dt><code>|matches| = |urlPattern|.{{URLPattern/test(input, baseURL)|test}}(|input|)</code></dt>
   <dd>
-    Tests if |urlPattern| matches the given arguments.  The |input| is an object containing strings representing each URL component; e.g. hostname, pathname, etc.  Missing components are treated as empty strings.  In addition, |input| can contain a baseURL property that provides values for any missing components.  If |urlPattern| matches the |input| on a component-by-component basis then true is returned.  Otherwise, false is returned.
+    Tests if |urlPattern| matches the given arguments.  The |input| is an object containing strings representing each URL component; e.g. hostname, pathname, etc.  Missing components are treated as empty strings.  In addition, |input| can contain a {{URLPatternInit/baseURL}} property that provides values for any missing components.  If |urlPattern| matches the |input| on a component-by-component basis then true is returned.  Otherwise, false is returned.
   </dd>
 
-  <dt><code>|matches| = |urlPattern|.{{URLPattern/test(input, baseURL)|test}}(|input|, |baseURL|)</code></dt>
+  <dt><code>|matches| = |urlPattern|.{{URLPattern/test(input, baseURL)|test}}(|url|, |baseURL|)</code></dt>
   <dd>
-    Tests if |urlPattern| matches the given arguments.  The |input| is a URL string.  If |baseURL| is provided, then |input| can be relative.  If |urlPattern| matches the |input| on a component-by-component basis then true is returned.  Otherwise, false is returned.
+    Tests if |urlPattern| matches the given arguments.  |url| is a URL string.  If |baseURL| is provided, then |url| can be relative.
+    
+    If |urlPattern| matches the |input| on a component-by-component basis then true is returned.  Otherwise, false is returned.
   </dd>
 
   <dt><code>|result| = |urlPattern|.{{URLPattern/exec(input, baseURL)|exec}}(|input|)</code></dt>
   <dd>
-    Executes the |urlPattern| against the given arguments.  The |input| is an object containing strings representing each URL component; e.g. hostname, pathname, etc.  Missing components are treated as empty strings.  In addition, |input| can contain a baseURL property that provides values for any missing components.  If |urlPattern| matches the |input| on a component-by-component basis then an object is returned containing the results.  Matched group values are contained in per-component group objects within the |result| object; e.g. `matches.pathname.group.id`.  If |urlPattern| does not match the |input|, then |result| is null.
+    Executes the |urlPattern| against the given arguments.  The |input| is an object containing strings representing each URL component; e.g. hostname, pathname, etc.  Missing components are treated as empty strings.  In addition, |input| can contain a baseURL property that provides values for any missing components.
+    
+    If |urlPattern| matches the |input| on a component-by-component basis then an object is returned containing the results.  Matched group values are contained in per-component group objects within the |result| object; e.g. `matches.pathname.groups.id`.  If |urlPattern| does not match the |input|, then |result| is null.
   </dd>
 
-  <dt><code>|result| = |urlPattern|.{{URLPattern/exec(input, baseURL)|exec}}(|input|, |baseURL|)</code></dt>
+  <dt><code>|result| = |urlPattern|.{{URLPattern/exec(input, baseURL)|exec}}(|url|, |baseURL|)</code></dt>
   <dd>
-    Executes the |urlPattern| against the given arguments.  The |input| is a URL string.  If |baseURL| is provided, then |input| can be relative..  If |urlPattern| matches the |input| on a component-by-component basis then an object is returned containing the results.  Matched group values are contained in per-component group objects within the |result| object; e.g. `matches.pathname.group.id`.  If |urlPattern| does not match the |input|, then |result| is null.
+    Executes the |urlPattern| against the given arguments.  |url| is a URL string.  If |baseURL| is provided, then |input| can be relative.
+    
+    If |urlPattern| matches the |input| on a component-by-component basis then an object is returned containing the results.  Matched group values are contained in per-component group objects within the |result| object; e.g. `matches.pathname.groups.id`.  If |urlPattern| does not match the |input|, then |result| is null.
   </dd>
 
   <dt><code>|urlPattern|.{{URLPattern/protocol}}</code></dt>
--- a/spec.bs
+++ b/spec.bs
@@ -186,44 +186,74 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
 Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component</dfn>, a [=component=], which must be set upon creation.
 
 <dl class="domintro non-normative">
-  <dt><code>{{URLPattern}} . {{URLPattern/protocol}}</code>
+  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL)|URLPattern}}(|input|)</code></dt>
   <dd>
-    <p>The normalized protocol pattern string.
+    Constructs a new URLPattern object.  The |input| is an object containing separate patterns for each URL component; e.g. hostname, pathname, etc.  Missing components will default to a wildcard pattern.  In addition, |input| can contain a baseURL property that provides static text patterns for any missing components.
   </dd>
 
-  <dt><code>{{URLPattern}} . {{URLPattern/username}}</code>
+  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL)|URLPattern}}(|input|, |baseURL|)</code></dt>
   <dd>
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic PTAL.  Thanks.

spec.bs
```diff
@@ -185,44 +185,59 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
 Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component</dfn>, a [=component=], which must be set upon creation.
 
 <dl class="domintro non-normative">
-  <dt><code>{{URLPattern}} . {{URLPattern/protocol}}</code>
+  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL)|URLPattern}}(|input|, |baseURL|)</code></dt>
+  <dd class=allow-2119>
+    Constructs a new URLPattern object.  The |input| may be a {{URLPatternInit}} dictionary or a string.  If |input| is a string, then an optional |baseURL| string may also be provided.
+  </dd>
```
domenic, line 191: I would suggest separate dt/dd pairs for the two overloads since they have rather different semantics and parameter orders. Even though they're not formal Web IDL overloads for the reasons discussed in #35, probably web developers should think of them as separate.

Same for test and exec.

spec.bs
```diff
@@ -185,44 +185,59 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
 Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component</dfn>, a [=component=], which must be set upon creation.
 
 <dl class="domintro non-normative">
-  <dt><code>{{URLPattern}} . {{URLPattern/protocol}}</code>
+  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL)|URLPattern}}(|input|, |baseURL|)</code></dt>
+  <dd class=allow-2119>
```
domenic, line 189: Turning off the 2119 checking is a smell. I'd suggest just saying "can" instead of "may".

spec.bs
```diff
@@ -185,44 +185,59 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
...
+    Constructs a new URLPattern object.  The |input| may be a {{URLPatternInit}} dictionary or a string.  If |input| is a string, then an optional |baseURL| string may also be provided.
```
domenic, line 190:
```suggestion
    Constructs a new {{URLPattern}} object.  The |input| may be a {{URLPatternInit}} dictionary or a string.  If |input| is a string, then an optional |baseURL| string may also be provided.
```

domenic: I prefer not to mention the concept "dictionary" (it's not something web developers are familiar with) or reference IDL constructs which don't have their own domintros. So I'd suggest something like "input can be an object containing protocol, username, password, hostname, port, pathname, search, hash, and baseURL values which are used to construct the pattern" for the constructor, and "which are matched against the pattern" for the others.

spec.bs
```diff
@@ -185,44 +185,59 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
...
+  </dd>
+
+  <dt><code>|matches| = |urlPattern|.{{URLPattern/test(input, baseURL)|test}}(|input|, |baseURL|)</code></dt>
+  <dd class=allow-2119>
+    Tests if |urlPattern| matches the given arguments.  The |input| may be a {{URLPatternInit}} dictionary or a string.  If |input| is a string, then an optional |baseURL| string may also be provided.  The returned |matches| value is a boolean.
+  </dd>
+
+  <dt><code>|result| = |urlPattern|.{{URLPattern/exec(input, baseURL)|exec}}(|input|, |baseURL|)</code></dt>
+  <dd class=allow-2119>
+    Executes the |urlPattern| against the given arguments.  The |input| may be a {{URLPatternInit}} dictionary or a string.  If |input| is a string, then an optional |baseURL| string may also be provided.  If the arguments did not match the |urlPattern|, then |result| will be null.  If the arguments did match, then |result| will be a {{URLPatternResult}} providing details about values that matched groups within |urlPattern|.
```
domenic, line 200: Similarly describing |result| in more detail than just referring to URLPatternResult would be ideal. https://html.spec.whatwg.org/#2dcontext:dom-context-2d-canvas-getcontextattributes-2 is an example of what we've done for return values in the past.

domenic (commented):
You don't have to take my advice about explaining all the dictionaries instead of linking to them, since it might bloat these into something fairly large. It's just my preference.

spec.bs
wanderview: I have been doing that for most of my notes.  I haven't been able to figure out consistently what terms I'm using are triggering it.  Honestly the checks are annoying and don't seem to add value to me.

wanderview: I started adding the exception after many iterations trying to change my phrasing and gave up.

domenic: The words are the words in https://infra.spec.whatwg.org/#conformance . It is pretty confusing for me as a reader to encounter those words in non-normative contexts, so I think it's worth trying to avoid bypassing the check when possible.

wanderview: The problem I run into is when I'm trying to describe things the web developer must do or we as the spec writer must algorithmically do, but they are not MUST statements directed at the implementor.  How do I phrase that?

And is it that confusing if you didn't notice the other dozen or so exceptions I already have in the doc?

domenic: You use the alternatives listed in https://infra.spec.whatwg.org/#conformance:

> For non-normative content "strongly encouraged", "strongly discouraged", "encouraged", "discouraged", "can", "cannot", "could", "could not", "might", and "might not" can be used instead. 

> And is it that confusing if you didn't notice the other dozen or so exceptions I already have in the doc?

This is indeed the first time it's really struck me, besides the last PR where allow-2119 seemed legitimate since it was a non-normative note explicitly discussing a lack of "required" normative requirements.

But it seems also pretty easy to fix in other places. E.g. "We found a protocol suffix, so this must be an absolute URLPattern constructor string" -> "We found a protocol suffix, so this **is** an absolute URLPattern constructor string".

spec.bs
wanderview: I tried to describe it better, but did not break it down like the example.  I ran out of time to try that today.  Also, since its got multiple nested objects, not sure if that would be very clear.

wanderview:
PTAL

spec.bs
```diff
@@ -186,44 +186,74 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
 Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component</dfn>, a [=component=], which must be set upon creation.
 
 <dl class="domintro non-normative">
-  <dt><code>{{URLPattern}} . {{URLPattern/protocol}}</code>
+  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL)|URLPattern}}(|input|)</code></dt>
   <dd>
-    <p>The normalized protocol pattern string.
+    Constructs a new URLPattern object.  The |input| is an object containing separate patterns for each URL component; e.g. hostname, pathname, etc.  Missing components will default to a wildcard pattern.  In addition, |input| can contain a baseURL property that provides static text patterns for any missing components.
```
domenic, line 191:
```suggestion
    Constructs a new {{URLPattern}} object.  The |input| is an object containing separate patterns for each URL component; e.g. hostname, pathname, etc.  Missing components will default to a wildcard pattern.  In addition, |input| can contain a {{URLPatternInit/baseURL}} property that provides static text patterns for any missing components.
```

(mainly just to get the nice code font)

spec.bs
```diff
@@ -186,44 +186,74 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
...
   </dd>
 
-  <dt><code>{{URLPattern}} . {{URLPattern/username}}</code>
+  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL)|URLPattern}}(|input|, |baseURL|)</code></dt>
```
domenic, line 194:
```suggestion
  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL)|URLPattern}}(|patternString|, |baseURL|)</code></dt>
```

spec.bs
```diff
@@ -186,44 +186,74 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
...
   <dd>
-    <p>The normalized username pattern string.
+    Constructs a new URLPattern object.  The |input| is a URL string containing pattern syntax for one or more components.  If |baseURL| is provided, then |input| can be relative.  This constructor will always set at least an empty string value and does not default any components to wildcard patterns.
```
domenic, line 196:
```suggestion
    Constructs a new {{URLPattern}} object.  |patternString| is a URL string containing pattern syntax for one or more components.  If |baseURL| is provided, then |patternString| can be relative.  This constructor will always set at least an empty string value and does not default any components to wildcard patterns.
```

spec.bs
```diff
@@ -186,44 +186,74 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
...
   </dd>
 
-  <dt><code>{{URLPattern}} . {{URLPattern/password}}</code>
+  <dt><code>|matches| = |urlPattern|.{{URLPattern/test(input, baseURL)|test}}(|input|)</code></dt>
   <dd>
-    <p>The normalized password pattern string.
+    Tests if |urlPattern| matches the given arguments.  The |input| is an object containing strings representing each URL component; e.g. hostname, pathname, etc.  Missing components are treated as empty strings.  In addition, |input| can contain a baseURL property that provides values for any missing components.  If |urlPattern| matches the |input| on a component-by-component basis then true is returned.  Otherwise, false is returned.
```
domenic, line 201:
```suggestion
    Tests if |urlPattern| matches the given arguments.  The |input| is an object containing strings representing each URL component; e.g. hostname, pathname, etc.  Missing components are treated as empty strings.  In addition, |input| can contain a {{URLPatternInit/baseURL}} property that provides values for any missing components.  If |urlPattern| matches the |input| on a component-by-component basis then true is returned.  Otherwise, false is returned.
```

spec.bs
```diff
@@ -186,44 +186,74 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
...
   </dd>
 
-  <dt><code>{{URLPattern}} . {{URLPattern/hostname}}</code>
+  <dt><code>|matches| = |urlPattern|.{{URLPattern/test(input, baseURL)|test}}(|input|, |baseURL|)</code></dt>
```
domenic, line 204:
```suggestion
  <dt><code>|matches| = |urlPattern|.{{URLPattern/test(input, baseURL)|test}}(|url|, |baseURL|)</code></dt>
```

spec.bs
```diff
@@ -186,44 +186,74 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
...
   <dd>
-    <p>The normalized hostname pattern string.
+    Tests if |urlPattern| matches the given arguments.  The |input| is a URL string.  If |baseURL| is provided, then |input| can be relative.  If |urlPattern| matches the |input| on a component-by-component basis then true is returned.  Otherwise, false is returned.
```
domenic, line 206:
```suggestion
    Tests if |urlPattern| matches the given arguments.  |url| is a URL string.  If |baseURL| is provided, then |url| can be relative.
    
    If |urlPattern| matches the |input| on a component-by-component basis then true is returned.  Otherwise, false is returned.
```

spec.bs
```diff
@@ -186,44 +186,74 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
...
   </dd>
 
-  <dt><code>{{URLPattern}} . {{URLPattern/port}}</code>
+  <dt><code>|result| = |urlPattern|.{{URLPattern/exec(input, baseURL)|exec}}(|input|)</code></dt>
   <dd>
-    <p>The normalized port pattern string.
+    Executes the |urlPattern| against the given arguments.  The |input| is an object containing strings representing each URL component; e.g. hostname, pathname, etc.  Missing components are treated as empty strings.  In addition, |input| can contain a baseURL property that provides values for any missing components.  If |urlPattern| matches the |input| on a component-by-component basis then an object is returned containing the results.  Matched group values are contained in per-component group objects within the |result| object; e.g. `matches.pathname.group.id`.  If |urlPattern| does not match the |input|, then |result| is null.
```
domenic, line 211:
```suggestion
    Executes the |urlPattern| against the given arguments.  The |input| is an object containing strings representing each URL component; e.g. hostname, pathname, etc.  Missing components are treated as empty strings.  In addition, |input| can contain a baseURL property that provides values for any missing components.
    
    If |urlPattern| matches the |input| on a component-by-component basis then an object is returned containing the results.  Matched group values are contained in per-component group objects within the |result| object; e.g. `matches.pathname.groups.id`.  If |urlPattern| does not match the |input|, then |result| is null.
```

spec.bs
```diff
@@ -186,44 +186,74 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
...
   </dd>
 
-  <dt><code>{{URLPattern}} . {{URLPattern/pathname}}</code>
+  <dt><code>|result| = |urlPattern|.{{URLPattern/exec(input, baseURL)|exec}}(|input|, |baseURL|)</code></dt>
```
domenic, line 214:
```suggestion
  <dt><code>|result| = |urlPattern|.{{URLPattern/exec(input, baseURL)|exec}}(|url|, |baseURL|)</code></dt>
```

spec.bs
```diff
@@ -186,44 +186,74 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
...
   <dd>
-    <p>The normalized pathname pattern string.
+    Executes the |urlPattern| against the given arguments.  The |input| is a URL string.  If |baseURL| is provided, then |input| can be relative..  If |urlPattern| matches the |input| on a component-by-component basis then an object is returned containing the results.  Matched group values are contained in per-component group objects within the |result| object; e.g. `matches.pathname.group.id`.  If |urlPattern| does not match the |input|, then |result| is null.
```
domenic, line 216:
```suggestion
    Executes the |urlPattern| against the given arguments.  |url| is a URL string.  If |baseURL| is provided, then |input| can be relative.
    
    If |urlPattern| matches the |input| on a component-by-component basis then an object is returned containing the results.  Matched group values are contained in per-component group objects within the |result| object; e.g. `matches.pathname.groups.id`.  If |urlPattern| does not match the |input|, then |result| is null.
```

domenic (approved):
Minor suggestions. I like the balance you struck of explaining without going into too-extensive detail.

PR #112, opened 2021-09-01
Replace 'should treat as standard URL' terminology. (Fixes #111)

### LINKED ISSUE #111: "should treat as a standard URL" is unclear
It's not clear what a standard URL is. "should" is also a conformance term and I don't think we should use it as part of other terms.

wanderview:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/112.html" title="Last updated on Sep 1, 2021, 6:50 PM UTC (bce1764)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/112/ab2794b...bce1764.html" title="Last updated on Sep 1, 2021, 6:50 PM UTC (bce1764)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -282,7 +282,7 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
   1. Set [=this=]'s [=URLPattern/password component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/password}}"], [=canonicalize a password=], and [=default options=].
   1. Set [=this=]'s [=URLPattern/hostname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/hostname}}"], [=canonicalize a hostname=], and [=hostname options=].
   1. Set [=this=]'s [=URLPattern/port component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/port}}"], [=canonicalize a port=], and [=default options=].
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic PTAL.

spec.bs
```diff
@@ -574,15 +574,15 @@ To <dfn>parse a constructor string</dfn> given a string |input|:
...
-            1. If |parser|'s [=constructor string parser/should treat as a standard URL=] is true, then set |parser|'s [=constructor string parser/result=]["{{URLPatternInit/pathname}}"] to "`/`".
+            1. Run [=compute protocol matches a special scheme flag=] given |parser|.
+              <p class="note allow-2119">We must eagerly compile the protocol component to determine if it matches any [=special schemes=].  If it does then certain special rules apply.  It determines if the pathname defaults to a "`/`" and also whether we should look for the username, password, hostname, and port components.  Authority slashes may also cause us to look for these components as well.  Otherwise we treat this as a "cannot be a base URL" and go straight to the pathname component.
```
domenic, line 578: It would be good to fix the allow-2119 while you're here. E.g. must -> need to, should -> will. Right now it's pretty confusing that you only "should" do those things, which implies an implementation might not do so.

domenic (commented):
Is "standard pathname" also problematic?

spec.bs
wanderview: I guess the URL spec just calls it "path" and "cannot-be-a-base-URL path".

domenic: I kind of like "pathname with slashes", but "can-be-a-base-URL pathname" might be more in line with URL spec terminology.

wanderview: I went with just "pathname" and "cannot-be-a-base-URL pathname" like the URL spec.  WDYT?

domenic (approved):
LGTM. "pathname" + "cannot-be-a-base-URL pathname" seems less clear than "can-be-a-base-URL pathname" + "cannot-be-a-base-URL pathname", but it's more symmetric with the URL Standard so yeah, we'd probably want to make any changes there first.

Issue #113, opened 2021-09-01
Support ipv6 addresses in constructor string


wanderview:
As pointed out in #110 its not possible to do `new URLPattern("http://[::1]/")`.  Even if you properly escape the colons like `new URLPattern("http://[\\:\\:1]/")` it still throws.

Adding support for this turns out to not be too hard.  The URL spec simply ignores colons inside `[ ]` brackets.  We can do the same in URLPattern.  Of course, for the hostname pattern to be valid the colon's will still need to be escaped.

PR #114, opened 2021-09-01
Add support for ipv6 hostnames to the constructor string parser. (Fix…

wanderview:
…es #113)

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/114.html" title="Last updated on Sep 1, 2021, 8:32 PM UTC (7715449)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/114/b091743...7715449.html" title="Last updated on Sep 1, 2021, 8:32 PM UTC (7715449)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -482,6 +482,8 @@ A [=constructor string parser=] has an associated <dfn export for="constructor s
 
 A [=constructor string parser=] has an associated <dfn export for="constructor string parser">group depth</dfn>, a number, initially set to 0.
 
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic PTAL.  This pretty much matches what the URL parser does to ignore colons while inside `[ ]` brackets:

https://url.spec.whatwg.org/#hostname-state

domenic (approved):
Pretty simple!

wanderview:
Actually, I wonder if the encoding callback to use the host parser in the url spec will be too restrictive.  Like, if it sees a partial pattern string '[::' will it throw because there is no enclosing bracket?  I'll check tomorrow.

PR #116, opened 2021-09-02
Support pattern syntax in ipv6 hostnames. (Fixes #115)

### LINKED ISSUE #115: cannot use pattern syntax inside ipv6 address
We added support to the constructor string parser over in #113.  This allows ipv6 literals like `new URLPattern("http://[\\:\\:1]")` to be used.  There is, however, a separate ipv6 problem.

If you have an ipv6 hostname it becomes impossible to use any pattern syntax inside the address.  For example, all of these currently throw:

```javascript
// these throw right now
new URLPattern({ hostname: '[:address]' });
new URLPattern("http://[:address]");
```

The problem is that we pass only the "plain text" parts of the pattern to the hostname encoding callback.  The callback then delegates to the URL parsing state machine.  So the URL parsing state machine sees a `[` string by itself.  This throws since the URL parsing state machine requires any hostname starting with `[` to also end with a `]` and then also successfully parse as ipv6.

We will need to use a more lenient hostname encoding callback for hostname patterns that appear to be ipv6.

wanderview:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/116.html" title="Last updated on Sep 3, 2021, 8:05 PM UTC (0b0c53a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/116/7971c0c...0b0c53a.html" title="Last updated on Sep 3, 2021, 8:05 PM UTC (0b0c53a)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -280,7 +280,8 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
   1. Set [=this=]'s [=URLPattern/protocol component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/protocol}}"], [=canonicalize a protocol=], and [=default options=].
   1. Set [=this=]'s [=URLPattern/username component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/username}}"], [=canonicalize a username=], and [=default options=].
   1. Set [=this=]'s [=URLPattern/password component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/password}}"], [=canonicalize a password=], and [=default options=].
-  1. Set [=this=]'s [=URLPattern/hostname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/hostname}}"], [=canonicalize a hostname=], and [=hostname options=].
+  1. If the result running [=hostname pattern is an IPv6 address=] given |processedInit|["{{URLPatternInit/hostname}}"] is true, then set [=this=]'s [=URLPattern/hostname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/hostname}}"], [=canonicalize an IPv6 hostname=], and [=hostname options=].
+  1. Else, set [=this=]'s [=URLPattern/hostname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/hostname}}"], [=canonicalize a hostname=], and [=hostname options=].
   1. Set [=this=]'s [=URLPattern/port component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/port}}"], [=canonicalize a port=], and [=default options=].
   1. If the result of running [=protocol component matches a special scheme=] given [=this=]'s [=URLPattern/protocol component=] is true, then set [=this=]'s [=URLPattern/pathname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/pathname}}"], [=canonicalize a pathname=], and [=pathname options=].
   1. Else set [=this=]'s [=URLPattern/pathname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/pathname}}"], [=canonicalize a cannot-be-a-base-URL pathname=], and [=default options=]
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic can you please review this PR with further ipv6 fixes?  I'm not sure I handled the code point/infra stuff right.  Thanks.

spec.bs
```diff
@@ -464,6 +465,16 @@ The <dfn>pathname options</dfn> is an [=options=] [=struct=] with [=options/deli
   1. Return false.
 </div>
 
+<div algorithm>
+  To determine if a <dfn>hostname pattern is an IPv6 address</dfn> given a [=/pattern string=] |input|:
+
+  1. If |input|'s [=string/code point length=] is less than 2, then return false.
+  1. If |input|[0] is U+005B (`[`), then return true.
+  1. If |input|[0] is U+007B (`{`) and |input|[1] is U+005B (`[`), then return true.
+  1. If |input|[0] is U+005C (<code>\</code>) and |input|[1] is U+005B (`[`), then return true.
```
domenic, line 474: So, the indexing syntax is not really well-defined for strings. If it were defined, it would probably refer to code units, not code points, since the primary definition is as a sequence of code units.

However, I think it's still OK and unambiguous to use here, because everything is ASCII. That is why I haven't mentioned it before. But since you asked for some focus on Unicode stuff in particular on this review, I thought I'd mention it.

If you really want to be cautious, probably the right phrasing would be something like "Let input0 be the 0th code point in input" and "Let input1 be the 1st code point in input", then work with those. (Although even there it's not well defined whether code points are 0-indexed or 1-indexed in such prose phrases. But by saying 0th we're making an implicit claim that they're 0-indexed, so it's pretty clear to the reader.)

/cc @annevk as an FYI about things to improve in Infra.

spec.bs
```diff
@@ -1522,6 +1533,23 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   1. Return |dummyURL|'s [=url/host=].
 </div>
 
+<div algorithm>
+  To <dfn>canonicalize an IPv6 hostname</dfn> given a string |value|:
+
+  1. Let |result| be the empty string.
+  1. [=list/For each=] |code point| in |value| interpreted as a sequence of [=/code points=]:
```
domenic, line 1540: I like this a lot.

spec.bs
```diff
@@ -1522,6 +1533,23 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+    1. If all of the following are true:
+      <ul>
+        <li>|code point| is not an [=ASCII hex digit=];</li>
+        <li>|code point| is not U+005B (`[`);</li>
+        <li>|code point| is not U+005D (`]`); and</li>
+        <li>|code point| is not U+003A (`:`),</li>
+      </ul>
+      <p>then throw a {{TypeError}}.
+    1. Append the result of running [=ASCII lowercase=] given "|code point|" to the end of |result|.
```
domenic, line 1549: Hmm. This "string literal with a variable" syntax is a bit weird; I thought you were using scare quotes.

A few ideas:

- Just use ASCII lowercase with an implicit conversion from code point to string. Infra can eventually define such an implicit conversion.
- Be very explicit, e.g. "Let _stringForCodePoint_ be a string consisting of the single code point _code point_" and then do that.
- Inline the definition of ASCII lowercase, i.e. "If _code point_ is an ASCII upper alpha, then append the corresponding code point in ASCII lower alpha to _result_. Otherwise, append _code point_ to _result_.

Probably the first is nicest.

domenic (commented):
This all looks great, but you have indeed found some areas where Infra does not give you a 100% rigorous foundation. In all cases though, it's pretty unambiguous what you mean, so it's mostly feedback for Infra and not something that prevents this PR from being merge-worthy.

spec.bs
wanderview: As discussed on chat, switch this to use:

`Let |input code points| be |input| interpreted as a [=list=] of [=/code points=].`

And then:

`|input code points|[0]`

spec.bs
wanderview: As per chat, changing "sequence" to "[=list=]".

wanderview:
Updated.  PTAL.

domenic (approved):

PR #129, opened 2021-09-06
Fix constructor string parser `change state` algo

lucacasonato:
Previously, the result was set based on "new state", not based on the existing state. This aligns to Chromium implementation: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/url_pattern/url_pattern_parser.cc;l=271-275;drc=7bd7edac514e6a13820d9982afaa8a5102bd7688

***
<a href="https://pr-preview.s3.amazonaws.com/lucacasonato/urlpattern/pull/129.html" title="Last updated on Sep 6, 2021, 2:44 PM UTC (5ac6eea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/129/7e8cb4c...lucacasonato:5ac6eea.html" title="Last updated on Sep 6, 2021, 2:44 PM UTC (5ac6eea)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -658,10 +658,10 @@ To <dfn>parse a constructor string</dfn> given a string |input|:
 </div>
 
 <div algorithm>
-To <dfn>change state</dfn> given a [=constructor string parser=] |parser|, a [=constructor string parser/state=] |state|, and a number |skip|:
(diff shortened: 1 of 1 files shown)


wanderview:
Good catch.  Thanks again.

PR #131, opened 2021-09-06
Handle no baseURL + relative input failure case

lucacasonato:
If the passed input to the URLPattern constructor is a relative path,
and no baseURL is set, throw a TypeError.

This behaviour aligns with Chromium implementation and tests: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/url_pattern/url_pattern.cc;l=213-219;drc=7bd7edac514e6a13820d9982afaa8a5102bd7688


wanderview:
Thanks!

Issue #145, opened 2021-10-26
pattern string canonicalization with wildcards can result in different meanings


wanderview:
Currently, when generating canonical pattern strings the algorithm always translates `.*` regex values into wildcards like `*`.  Unfortunately, this can result in a change of meaning if it does not consider the values before the regexp group.

Consider, this:

  `(foo)(.*)`

Gets "canonicalized" to:

  `(foo)*`

If this is parsed again the `*` is treated as a modifier of `(foo)` instead of as a wildcard.

The algorithm should be fixed to look at the preceding part to determine if its safe to convert a `.*` regexp into a wildcard character.

wanderview:
Another type of problem here with:

`{:foo}(.*)`

Assuming we fix this not to convert `(.*)` to `*`, we also need to avoid accidentally changing it to:

`:foo(.*)`

That changes the pattern from two groups to a single group with a custom name and wildcard regexp.

wanderview:
Another slight variation on the problem:

`{:foo}bar`

Becomes:

`:foobar`

lucacasonato:
@wanderview The WPTs seem to have landed for this, and the Chromium change also seems to have landed, but there is no spec change yet. Is this in the works?

lucacasonato:
Ah no, it is #152. Sorry.

wanderview:
Sorry, I went on holiday prior to finishing the spec changes.  It will be the first thing I do now that I'm back.

wanderview:
I believe this is fixed in the spec now.  I will work on updating the polyfill.

wanderview:
We found another issue along these lines:

`/foo\\/:bar?`

Should require the second slash because the escape prevents it from becoming a prefix to the `:bar?` group.  Pattern string generation, however, produces this:

`/foo/:bar?`

This was found by our fuzzer here:

https://bugs.chromium.org/p/chromium/issues/detail?id=1285620

wanderview:
Another one of these was found by the fuzzer:

https://bugs.chromium.org/p/chromium/issues/detail?id=1289115

In this case if an implicit prefix is before `:foo` group it can incorrectly prevent us from protecting from trailing fixed characters.

PR #151, opened 2021-12-09
Don't emit '*' when it can be confused for a modifier. (#145)

wanderview:
This fixes one of the cases outlined in #145.  When we have a matching group
following by a wildcard we must be careful not to emit `*` for the wildcard.
We don't want it to be mistaken for modifier on the previous group.  So we
want to generate `(foo)(.*)` and not `(foo)*`.

This change corresponds to this implementation change in chromium:

https://chromium-review.googlesource.com/c/chromium/src/+/3313642

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/151.html" title="Last updated on Dec 9, 2021, 9:10 PM UTC (6f57c26)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/151/dbb9521...6f57c26.html" title="Last updated on Dec 9, 2021, 9:10 PM UTC (6f57c26)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/151.html" title="Last updated on Jan 5, 2022, 4:26 PM UTC (611d26f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/151/dbb9521...611d26f.html" title="Last updated on Jan 5, 2022, 4:26 PM UTC (611d26f)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1421,7 +1421,10 @@ To <dfn>escape a regexp string</dfn> given a string |input|:
 To <dfn export>generate a [=/pattern string=]</dfn> from a given [=/part list=] |part list| and [=/options=] |options|:
 
   1. Let |result| be the empty string.
-  1. [=list/For each=] |part| of |part list|:
+  1. Let |index list| be the result of [=list/getting the indices=] for |part list|.
+  1. [=list/For each=] |index| of |index list|:
+    1. Let |part| be |part list|[|index|].
+    1. Let |last part| be |part list|[|index| - 1] if |index| is greater than 0, otherwise let it be null.
     1. If |part|'s [=part/type=] is "<a for=part/type>`fixed-text`</a>" then:
       1. If |part|'s [=part/modifier=] is "<a for=part/modifier>`none`</a>" then:
         1. Append the result of running [=escape a pattern string=] given |part|'s [=part/value=] to the end of |result|.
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic PTAL

spec.bs
```diff
@@ -1421,7 +1421,10 @@ To <dfn>escape a regexp string</dfn> given a string |input|:
 To <dfn export>generate a [=/pattern string=]</dfn> from a given [=/part list=] |part list| and [=/options=] |options|:
 
   1. Let |result| be the empty string.
-  1. [=list/For each=] |part| of |part list|:
+  1. Let |index list| be the result of [=list/getting the indices=] for |part list|.
+  1. [=list/For each=] |index| of |index list|:
+    1. Let |part| be |part list|[|index|].
+    1. Let |last part| be |part list|[|index| - 1] if |index| is greater than 0, otherwise let it be null.
```
domenic, line 1427: You could move this under the "Else if full-wildcard"

spec.bs
```diff
@@ -1452,7 +1455,12 @@ To <dfn export>generate a [=/pattern string=]</dfn> from a given [=/part list=]
       1. Append the result of running [=generate a segment wildcard regexp=] given |options| to the end of |result|.
       1. Append "`)`" to the end of |result|.
     1. Else if |part|'s [=part/type=] is "<a for=part/type>`full-wildcard`</a>":
-      1. If |custom name| is true:
+      1. If one of the following are true:
+        <ul>
+          <li>|custom name| is true.</li>
```
domenic, line 1460: Specs are very inconsistent but my preferred style these days for this sort of thing is

> If one of the following is true:
> 
> - x; or
> - y
> 
> then:

domenic (approved):
LGTM with optional editorial nits

spec.bs
domenic: Oh, I see, that would be less nice in combination with #152.

Separate issue though: "last" is confusing; it might mean the last one in the list. I'd suggest "previous"

wanderview:
Thanks.  I will fix this up after the holiday next year.

PR #152, opened 2021-12-09
Emit `{}` around custom names followed by a regexp group. (#145)

wanderview:
This change fixes another case from #145.  Here we need protect a `:foo`
group with `{}` braces if its followed by a matching group that may be
expressed as `(bar)`.  Without this protection the second group may be
interpreted as a custom regexp for the first group.  We want to generate
`{:foo}(bar)` and not `:foo(bar)`.

For the most part this change corresponds to this chromium commit:

https://chromium-review.googlesource.com/c/chromium/src/+/3315419

Note, however, this spec change includes an additional fix to check for
the next part being a segment wildcard.  That additional check will need
to be added to the implementation in another change.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/152.html" title="Last updated on Dec 9, 2021, 10:06 PM UTC (561f6cf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/152/6f57c26...561f6cf.html" title="Last updated on Dec 9, 2021, 10:06 PM UTC (561f6cf)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/152.html" title="Last updated on Jan 5, 2022, 3:50 PM UTC (101f35c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/152/6f57c26...101f35c.html" title="Last updated on Jan 5, 2022, 3:50 PM UTC (101f35c)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1425,7 +1425,7 @@ To <dfn export>generate a [=/pattern string=]</dfn> from a given [=/part list=] 
   1. [=list/For each=] |index| of |index list|:
     1. Let |part| be |part list|[|index|].
     1. Let |last part| be |part list|[|index| - 1] if |index| is greater than 0, otherwise let it be null.
-    1. Let |next part| be |part list|[|index| + 1] if |index| is less than |index list|'s [=list/size] = 1, otherwise let it be null.
+    1. Let |next part| be |part list|[|index| + 1] if |index| is less than |index list|'s [=list/size] - 1, otherwise let it be null.
     1. If |part|'s [=part/type=] is "<a for=part/type>`fixed-text`</a>" then:
       1. If |part|'s [=part/modifier=] is "<a for=part/modifier>`none`</a>" then:
         1. Append the result of running [=escape a pattern string=] given |part|'s [=part/value=] to the end of |result|.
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic PTAL.  Note, this depends on the previous pull request.  Thanks.

spec.bs
```diff
@@ -1425,6 +1425,7 @@ To <dfn export>generate a [=/pattern string=]</dfn> from a given [=/part list=]
   1. [=list/For each=] |index| of |index list|:
     1. Let |part| be |part list|[|index|].
     1. Let |last part| be |part list|[|index| - 1] if |index| is greater than 0, otherwise let it be null.
+    1. Let |next part| be |part list|[|index| + 1] if |index| is less than |index list|'s [=list/size] = 1, otherwise let it be null.
```
domenic, line 1428:
```suggestion
    1. Let |next part| be |part list|[|index| + 1] if |index| is less than |index list|'s [=list/size] - 1, otherwise let it be null.
```

spec.bs
```diff
@@ -1434,13 +1435,14 @@ To <dfn export>generate a [=/pattern string=]</dfn> from a given [=/part list=]
       1. Append "`}`" to the end of |result|.
       1. Append the result of running [=convert a modifier to a string=] given |part|'s [=part/modifier=] to the end of |result|.
       1. [=Continue=].
+    1. Let |custom name| be true if |part|'s [=part/name=][0] is not an [=ASCII digit=].
```
domenic, line 1438:
```suggestion
    1. Let |custom name| be true if |part|'s [=part/name=][0] is not an [=ASCII digit=]; otherwise false.
```

domenic (approved):
LGTM with nits

PR #153, opened 2022-01-05
Escape some suffixes trailing `:foo` groups. (#145)

wanderview:
This fixes one of the cases outlined in #145.  When a named group, like
`:foo`, has a suffix that begins with a valid name character, like `bar`,
we run the risk of generating a string that adds the suffix to the name
itself and not as a suffix to the group.  To avoid this we insert an
escape character before the suffix.  For example, we want to generate
`:foo\bar` and not `:foobar`.

This spec change corresponds to this implementation change:

https://chromium-review.googlesource.com/c/chromium/src/+/3318605

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/153.html" title="Last updated on Jan 6, 2022, 2:59 PM UTC (cbd5dfa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/153/3c8a799...cbd5dfa.html" title="Last updated on Jan 6, 2022, 2:59 PM UTC (cbd5dfa)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1467,6 +1467,14 @@ To <dfn export>generate a [=/pattern string=]</dfn> from a given [=/part list=] 
         1. Append [=full wildcard regexp value=] to the end of |result|.
         1. Append "`)`" to the end of |result|.
       1. Else append "`*`" to the end of |result|.
+    1. If all of the following are true:
+      <ul>
+        <li>|part|'s [=part/type=] is "<a for=part/type>`segment-wildcard`</a>"; and</li>
+        <li>|custom name| is true; and</li>
+        <li>|part|'s [=part/suffix=] is not the empty string; and</li>
+        <li>The result of running [=is a valid name code point=] given |part|'s [=part/suffix=]'s first [=/code point=] and the boolean false is true</li>
+      </ul>
+      then append U+005C (<code>\</code>) to the end of |result|.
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic PTAL.  Thanks.

domenic (approved):

spec.bs
```diff
@@ -1467,6 +1467,14 @@ To <dfn export>generate a [=/pattern string=]</dfn> from a given [=/part list=]
         1. Append [=full wildcard regexp value=] to the end of |result|.
         1. Append "`)`" to the end of |result|.
       1. Else append "`*`" to the end of |result|.
+    1. If all of the following are true:
+      <ul>
+        <li>|part|'s [=part/type=] is "<a for=part/type>`segment-wildcard`</a>"; and</li>
+        <li>|custom name| is true; and</li>
+        <li>|part|'s [=part/suffix=] is not the empty string; and</li>
+        <li>The result of running [=is a valid name code point=] given |part|'s [=part/suffix=]'s first [=/code point=] and the boolean true is true</li>
```
domenic, line 1475: Wait, the implementation uses false here instead of true, is that OK?

wanderview: Good catch.  Switching to false.

PR #155, opened 2022-01-10
URLPattern: Emit '{ }' when there is a preceding non-prefix '/'. (#145)

wanderview:
This commit fixes another issue from #145.

In this case we want to make sure that a character that can be treated
as an implicit group prefix, e.g. `/`, is not incorrectly converted into
a prefix.  For example, if the original string is `/{:foo?}` we don't
want to generate `/:foo?` since that would make the `/` optional.

This CL also includes an additional fix that permits emitting `*` for
wildcards when there is a prefix.  Previously we were being too
conservative in some situations.  For example, we can emit `*/*` instead
of `*/(.*)`.

This commit corresponds to this implementation change:

https://chromium-review.googlesource.com/c/chromium/src/+/3378528

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/155.html" title="Last updated on Jan 11, 2022, 4:35 PM UTC (dd8a1a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/155/9166150...dd8a1a5.html" title="Last updated on Jan 11, 2022, 4:35 PM UTC (dd8a1a5)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1444,6 +1444,7 @@ To <dfn export>generate a [=/pattern string=]</dfn> from a given [=/part list=] 
     1. If all of the following are true:
       <ul>
         <li>|needs grouping| is false; and</li>
+        <li>|part|'s [=part/prefix=] is the empty string; and</li>
         <li>|custom name| is true; and</li>
         <li>|part|'s [=part/type=] is "<a for=part/type>`segment-wildcard`</a>"; and</li>
         <li>|part|'s [=part/modifier=] is "<a for=part/modifier>`none`</a>"; and</li>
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic PTAL.  This is for another string generation issue found by the fuzzer.  The implementation CL has not landed yet.

domenic (commented):
I don't see any counterpart to the changes at line 212-216 in the implementation, and in general those lines don't seem to correspond as well as I'd hope? Is that OK?

wanderview:
Good catch!  I refactored the spec logic to look more like the implementation and added the missing changes.  PTAL.

domenic (approved):

Issue #156, opened 2022-01-17
pathname `.` collapsing triggering too much


crowlKats:
test code:
```js
const p = new URLPattern({ pathname: '/:foo.' });
console.log(p.test({pathname: "/bar."}));
console.log(p.test({pathname: "/bar"}));
```
Chrome:
```
true
false
```
urlpattern-polyfill:
```
true
true
```
deno currently panics, but with a quick fix for the panic, i reach same result as urlpattern-polyfill. Is something missing in the spec?

wanderview:
Hmm, I patched the immediate issue, but I'm not sure its adequate now.

Consider `new URLPattern({ pathname: ':foo./' })`.  Per the spec and the polyfill this gets converted to just `:foo` which seems wrong.  Chromium preserves it as `:foo./` instead.

On the flip side, though, what would we expect for `new URLPattern({ pathname: './foo' })`.  Note there is no base URL here.  Chromium keeps it as `./foo`, but the spec and polyfill collapse the `./` to just `foo`.  I think the chromium behavior is more correct here, but not as sure.  (Note, if there is a base URL with a pathname ending in a slash, then chromium does collapse correctly.)

I need to think about what the correct fix is here.

wanderview:
I think maybe the solution could be pre-pending a prefix like "/-" before we pass a value to the URL pathname parser if the value does not already start with a slash.  Then we would strip it back off at the end.

domenic:
I think this could benefit from some discussion with users, if possible. It kind of depends how much people expect these sort of strings to be "path-like" (where `./foo` is equivalent to `foo`) or "regexp-like", and how inserting `:bar` in various locations changes their intuition. I don't have much of a handle on that...

wanderview:
I'm not sure users will have a strong informed opinion here.  Its a bit esoteric.

Also, there is a rationale for the chromium behavior.  We only perform `.` and `..` collapsing when we are certain there are slashes on both sides.

The issue with the non-chromium behavior is an artifact of how we have to use the URL parser algorithm that implicitly inserts a leading `/` because its resolving full URLs.  We basically don't want that behavior here.

I also think keeping `new URLPattern({ pathname: "./foo" })` as `./foo` is principled because it has not been resolved against a base URL.  Its an unresolved path and so the leading `./` does not meet the criteria for collapsing.

Having thought this through more I feel pretty strongly that we should stick with the chromium behavior.

wanderview:
[...]

PR #157, opened 2022-01-18
Fix pathname encoding for isolated '.' and '..' values. (#156)

wanderview:
This commit fixes an issue where '.' and '..' could be incorrectly
removed from a pathname during parsing.  This would happen in cases
where the encoding callback saw the '.' or '..' individually without
surrounding characters.  This would cause the algorithm to encode
them as '/.' or '/..' which the URL parser collapses to just '/'.

To avoid this behavior we explicitly check for '.' and '..' values
so they can be immediately returned without encoding.

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/157.html" title="Last updated on Jan 18, 2022, 8:20 PM UTC (defa04d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/157/7336960...defa04d.html" title="Last updated on Jan 18, 2022, 8:20 PM UTC (defa04d)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1618,7 +1618,13 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
 <div algorithm>
   To <dfn>canonicalize a pathname</dfn> given a string |value|:
 
-  1. If |value| is the empty string, return |value|.
+  1. If |value| is one of the following:
+    <ul>
+      <li>the empty string; or</li>
+      <li>"`.`"; or</li>
+      <li>"`..`"
+    </ul>
+    then return |value|.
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic PTAL.  Thanks.

domenic (approved):
This LGTM, although I'm not 100% confident since the Chromium implementation doesn't seem to match the spec (it delegates to existing URL util code). Hopefully the Deno folks, who seem to be fast-following with a more spec-exact implementation, can confirm that this does the trick.

wanderview:
The polyfill is closer to the spec and this change fixes the problem there.

PR #166, opened 2022-05-03
Add `URLPatternList`

### LINKED ISSUE #30: should we still expose URLPatternList?
After the last face-to-face meeting we changed the plan to make service worker scope take a pathname patter string instead of a full URLPattern object and an array of strings instead of a URLPatternList object.

Given this was the main use case for URLPatternList, should we just not expose URLPatternList?  Are there other use cases for URLPatternList?

@jeffposnick @jakearchibald do you think developers would need URLPatternList vs simply just managing a list of URLPattern objects themselves?

lucacasonato:
This commit adds support for a `URLPatternList` class.

TODO:
- [ ] consensus on sort order
- [ ] check that a radix tree implementation is possible (implement in Deno)
- [ ] many many WPTs

Closes #30

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -810,6 +810,83 @@ To <dfn>compute protocol matches a special scheme flag</dfn> given a [=construct
   1. If the result of running [=protocol component matches a special scheme=] given |protocol component| is true, then set |parser|'s [=constructor string parser/protocol matches a special scheme flag=] to true.
 </div>
 
+<h2 id=urlpatternlist-class>The {{URLPatternList}} class </h2>
+
+<xmp class="idl">
+[Exposed=(Window,Worker)]
+interface URLPatternList {
+  constructor(sequence<URLPatternListEntry> entries);
+
+  USVString? test(optional URLPatternInput input = {}, optional USVString baseURL);
+
+  URLPatternListResult? exec(optional URLPatternInput input = {}, optional USVString baseURL);
+};
+
+dictionary URLPatternListEntry {
+  required (USVString or URLPattern) pattern;
+  required USVString key;
+};
+
+dictionary URLPatternListResult {
+  URLPatternResult match;
+  USVString key;
+};
+</xmp>
+
+Note: The {{URLPatternList}} is a utility that combines many patterns into one
+matcher. This functionality can also be achieved in a naive userland
+implementation that iterates over a list of patterns and matches each one
+individually. This has the unfortunate effect of not being very fast. With the
+built-in {{URLPatternList}} implementers can use a radix tree for matching under
+the hood which can drastically improve performance over the naive userland
+implementation.
+
+Each {{URLPatternList}} has an associated <dfn for=URLPatternList>pattern list</dfn>, a [=list=] of {{URLPattern}}s.
+
+Each {{URLPatternList}} has an associated <dfn for=URLPatternList>key list</dfn>, a [=list=] of [=string=]s.
+
+<div algorithm>
+  The <dfn constructor for=URLPatternList lt="URLPatternList(entries)">new URLPattern(|entries|)</dfn> constructor steps are:
+
+  1. [=list/For each=] |entry| in |entries|, run the following steps:
+    1. [=list/Append=] the |entry|'s {{URLPatternListEntry/pattern}} to [=this=]'s [=URLPatternList/pattern list=].
+    1. Let |key| be |entry|'s {{URLPatternListEntry/key}}.
+    1. If |key| is the empty string, throw a {{TypeError}}.
+    1. If [=this=]'s [=URLPatternList/key list=] [=list/contains=] |key|, throw a {{TypeError}}.
+    1. [=list/Append=] |key| to [=this=]'s [=URLPatternList/key list=].
+  1. If [=this=]'s [=URLPatternList/pattern list=] is empty, throw a {{TypeError}}.
+
+  Issue: The patterns need to be in a "least -> most significant" sort order so
+  implementers can efficiently use a radix tree under the hood. Enforcing /
+  sorting the patterns in this order depends on
+  <a href=https://github.com/WICG/urlpattern/issues/61>https://github.com/WICG/urlpattern/issues/61</a>.
+</div>
+
+<div>
+  The <dfn method for=URLPatternList>test(|input|, |baseURL|)</dfn> method steps are:
+
+  1. Let |i| be 0.
+  1. [=list/For each=] |pattern| in [=this=]'s [=URLPatternList/pattern list=], run the following steps:
+    1. Let |result| be the result of [=match=] given |pattern|, |input|, and |baseURL| if given.
+    1. If |result| is null, [=continue=].
+    1. Return [=this=]'s [=URLPatternList/key list=][|i|].
+  1. Return null.
+</div>
+
+<div>
+  The <dfn method for=URLPatternList>match(|input|, |baseURL|)</dfn> method steps are:
+
+  1. Let |i| be 0.
+  1. [=list/For each=] |pattern| in [=this=]'s [=URLPatternList/pattern list=], run the following steps:
+    1. Let |match| be the result of [=match=] given |pattern|, |input|, and |baseURL| if given.
+    1. If |match| is null, [=continue=].
+    1. Let |result| be a new {{URLPatternListResult}} dictionary.
+    1. Set |result|'s {{URLPatternListResult/match}} to |match|.
+    1. Set |result|'s {{URLPatternListResult/key}} to [=this=]'s [=URLPatternList/key list=][|i|].
+    1. Return |result|.
+  1. Return null.
+</div>
+
 <h2 id=patterns>Patterns</h2>
 
 A <dfn>pattern string</dfn> is a string that is written to match a set of target strings.  A <dfn for="pattern string">well formed</dfn> pattern string conforms to a particular pattern syntax.  This pattern syntax is directly based on the syntax used by the popular [path-to-regexp](https://github.com/pillarjs/path-to-regexp) JavaScript library.


spec.bs
```diff
@@ -810,6 +810,83 @@ To <dfn>compute protocol matches a special scheme flag</dfn> given a [=construct
...
+  URLPatternResult match;
+  USVString key;
+};
+</xmp>
+
+Note: The {{URLPatternList}} is a utility that combines many patterns into one
+matcher. This functionality can also be achieved in a naive userland
+implementation that iterates over a list of patterns and matches each one
+individually. This has the unfortunate effect of not being very fast. With the
+built-in {{URLPatternList}} implementers can use a radix tree for matching under
```
wanderview, line 840: Recommend just mentioning optimizations and not "radix tree" specifically.  That is just one kind of optimization.

spec.bs
```diff
@@ -810,6 +810,83 @@ To <dfn>compute protocol matches a special scheme flag</dfn> given a [=construct
...
+  constructor(sequence<URLPatternListEntry> entries);
+
+  USVString? test(optional URLPatternInput input = {}, optional USVString baseURL);
+
+  URLPatternListResult? exec(optional URLPatternInput input = {}, optional USVString baseURL);
+};
+
+dictionary URLPatternListEntry {
+  required (USVString or URLPattern) pattern;
+  required USVString key;
```
wanderview, line 827: Can you explain the "key" a bit more?  It feels like its not really functioning as a key here and is instead some extra data associated with the pattern.  It seems the pattern is more like the "key" of the entry to me.

Also, why a string for the extra data?  Why not an `any` so you can store any javascript object?

spec.bs
```diff
@@ -810,6 +810,83 @@ To <dfn>compute protocol matches a special scheme flag</dfn> given a [=construct
   1. If the result of running [=protocol component matches a special scheme=] given |protocol component| is true, then set |parser|'s [=constructor string parser/protocol matches a special scheme flag=] to true.
 </div>
 
+<h2 id=urlpatternlist-class>The {{URLPatternList}} class </h2>
+
+<xmp class="idl">
+[Exposed=(Window,Worker)]
+interface URLPatternList {
+  constructor(sequence<URLPatternListEntry> entries);
```
wanderview, line 818: I think some of the feedback in https://github.com/WICG/urlpattern/issues/30 suggested it would be nice to be able to dynamically mutate the list.  So add new entries to the list and possibly delete entries from the list.

wanderview (commented):
Thanks for looking at this!

wanderview:
Also:

> check that a radix tree implementation is possible (implement in Deno)

I doubt the API shape here will really dictate whether a radix tree implementation is possible or not.  Instead I expect the path-to-regexp style pattern matching will impose some constraints which make a pure radix solution difficult.

That does not mean optimizations will not be possible, though.  I think there are clear optimizations a list object like this can enable; e.g. log(n) searching of a fixed string prefix in the patterns.

I have also theorized it might be possible to construct an optimization where the patterns are deconstructed into automata similar to regex compilation.  Then something akin to a radix tree could be built from the automata.  We could then evaluate the automata against the input in O(n) time where n is the input length.

spec.bs
lucacasonato: No real reason. I think we may even be able to get rid of this "key" or metadata thing entirely. Instead of returning the `"key"` from the matcher, we can return the `URLPattern` object that matched. Users can associate metadata with a pattern using a `Map` or `WeakMap`. That way the input can just be `sequence<URLPattern>`.

spec.bs
lucacasonato: Makes sense. I didn't want to add this right away until we figure out the internal sort order a little better.

lucacasonato:
Yeah, that makes total sense. My first pass was going to just deal with speeding up fixed text prefix matches (actually also for regular `URLPattern`).

> I have also theorized it might be possible to construct an optimization where the patterns are deconstructed into automata similar to regex compilation. Then something akin to a radix tree could be built from the automata. We could then evaluate the automata against the input in O(n) time where n is the input length.

I was also thinking about that today! :)

I was actually planning to do exactly this for a v2 of the `URLPattern(List)` implementation in Deno. The way I envision it is to put another intermediate layer in between the "parts" and the "regexp" representation of a component. This intermediate format would be a layered "matcher" stack that is flat for `URLPattern`, and a proper tree for `URLPatternList`. You can then walk through this tree similar to a radix tree (except that some amount of backtracing is likely necessary). Each node in the tree would be some specific match construct. For example the simplest node would just be a literal string match. A more complex "optional" node would have an inner matcher that it would attempt to match before passing off to the next matcher in the stack. There would be a "regexp" node that just delegates a given match to the ECMA regexp engine (for user provided regexps).

What do you think about including this intermediate format in the spec? Doing that would allow us to make sure future spec additions / changes don't break non-regexp matcher algorithms. The spec would still only specify the naive regexp based matcher, but it'd be much easier for implementations to implement alternative faster matchers based on this intermediate format.

wanderview:
> What do you think about including this intermediate format in the spec?

I don't think we should put any intermediate representation in the spec as it then makes future optimizations more difficult in the implementation.  I think we want the flexibility to change the intermediate representation in the future.  Also, none of this stuff should be observable to sites.

wanderview:
Also, when this is further along it would be nice to include a "mini explainer" in the pull request description.  Something like https://github.com/whatwg/dom/pull/1032.  And submit to TAG review.

Besides helping to get feedback on the proposal this would also make it easier for chromium to adopt the API more quickly.

lucacasonato:
@wanderview I simplified the proposed API. There is no more `key` anymore. Instead, the `exec` method now returns a `pattern` field in the result object that users can use to figure out which pattern matched. To do attach a key (or any metadata) to a pattern, users can now use a `Map`:

```ts
const patterns: Array<[URLPattern, string]>;

const keys = new Map(patterns);
const list = new URLPatternList(patterns.map(r => r[0]));

const res = list.exec(...)
if (res !== null) console.log("Matched", map.get(res.pattern));
```

wanderview:
I think @domenic had previously advocated the map approach to me.  If this is ergonomic enough for developers, that works for me.

lucacasonato:
I have added a mini explainer to the description.

wanderview:
@lucacasonato What is the status of this?  How did your prototyping in deno go?  I'm thinking of prototyping this in chromium using re2's set implementation.

wanderview:
I was also wondering if we should rename this to `URLPatternSet`.  My impression is we don't want to support duplicate entries.

wanderview:
I filed an intent to prototype for chromium:

https://groups.google.com/a/chromium.org/g/blink-dev/c/QrPrveVyFnA/m/PkaPKfJ4AwAJ

wanderview:
Also, I wonder if the Map oriented approach to associating data with entries will make it hard to have a constructor that takes dictionaries instead of full URLPatterns.  Not sure if the dictionaries shortcut is worth optimizing for at the cost of added complexity elsewhere.

domenic:
I've always been -1 on the map-oriented approach, myself.

Regarding setlike APIs, I think the main question is whether you want people to introspect and possibly mutate these things after creation. The spec PR here gives an immutable, non-introspectable object, and that seems to accomplish the main use cases. On the other hand, add introspection APIs from setlike (entries, forEach, has, keys, size, values) doesn't seem to hurt. And adding add, clear, and delete also doesn't seem that bad.

Note that if you're not using insertion order, you'll need to override the implementation of add(), in order to put the given value in the right place in the [set entries](https://webidl.spec.whatwg.org/#dfn-set-entries).

wanderview:
Ok, I recall you were positive on passing in an array of init dictionaries to the constructor.  But that does seem to work with the external map approach.  Does that seem ok to you?

domenic:
Yeah, that seems OK to me. I think people who want to associate data that way can use explicit URLPattern objects.

spec.bs
```diff
@@ -810,6 +810,68 @@ To <dfn>compute protocol matches a special scheme flag</dfn> given a [=construct
...
+interface URLPatternList {
+  constructor(sequence<URLPattern> patterns);
+
+  boolean test(optional URLPatternInput input = {}, optional USVString baseURL);
+
+  URLPatternListResult? exec(optional URLPatternInput input = {}, optional USVString baseURL);
+};
+
+dictionary URLPatternListResult : URLPatternResult {
+  URLPattern pattern;
```
posva, line 826: I don't know how it works internally but shouldn't it contain the match result as well? Since the point is to pick a pattern to match and it's likely doing some work behind the scenes

spec.bs
```diff
@@ -810,6 +810,68 @@ To <dfn>compute protocol matches a special scheme flag</dfn> given a [=construct
...
+built-in {{URLPatternList}} implementers can use an optimized implementation for
+matching under the hood which can drastically improve performance over the naive
+userland implementation.
+
+Each {{URLPatternList}} has an associated <dfn for=URLPatternList>pattern list</dfn>, a [=list=] of {{URLPattern}}s.
+
+<div algorithm>
+  The <dfn constructor for=URLPatternList lt="URLPatternList(patterns)">new URLPatternList(|patterns|)</dfn> constructor steps are:
+
+  1. [=list/For each=] |pattern| in |patterns|, run the following steps:
```
posva, line 843: Doesn't this imply that the pattern list is traversed as an array and is potentially slower? https://github.com/WICG/urlpattern/issues/30#issuecomment-1089229462

spec.bs
bathos: It does — the `: URLPatternResult` part means this dictionary extends [URLPatternResult](https://wicg.github.io/urlpattern/#dictdef-urlpatternresult).

spec.bs
bathos: The iteration steps here and elsewhere don’t appear to be observable, so engines can potentially optimize them however they see fit, right? Spec steps usually aim for the simplest explanation of observable behaviors (because observable behaviors are the only things they _can_ actually specify, further complexity is noise).

spec.bs
posva: Aaah i didn’t see it. (Not familiar with this spec language 😂)

spec.bs
posva: I see, thanks for the explanation!

TomokiMiyauci:
Does this only do stateless longest matching?

Should it be possible to resume pattern matching from a specified pattern (`lastPattern`), like `RegExp`'s `lastIndex`?

mcollina:
@lucacasonato what's the status of this?

EisenbergEffect:
I'd love to see this happen. Any progress?

PR #168, opened 2022-07-13
feat: support for case-insensitive matches

### LINKED ISSUE #148: No option to make URLPattern case insensitive
There should be an option to make the test case insensitive. Right now it's always case sensitive:

```js
const p = new URLPattern('/A/:id', 'http://e.e')
p.exec('/A/2', 'http://e.e') // { ... }
p.exec('/a/2', 'http://e.e') // null
```

Sayan751:
This commit updates the specification document with the support of
case-insensitive matches by introducing a third argument in the
`URLPattern` constructor. This additional document contains a single
boolean property named `ignoreCase`, which can be set to `true` to
enable case-insensitive matches. By default, this flag is set to `false`,
thus enabling case-sensitive matches.

Closes #148

***
<a href="https://pr-preview.s3.amazonaws.com/Sayan751/urlpattern/pull/168.html" title="Last updated on Sep 22, 2022, 8:37 PM UTC (a53f5ca)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/168/26b794d...Sayan751:a53f5ca.html" title="Last updated on Sep 22, 2022, 8:37 PM UTC (a53f5ca)">Diff</a>


spec.bs
```diff
@@ -184,16 +189,28 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
...
   </dd>
 
+  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL, options)|URLPattern}}(|input|, |options|)</code></dt>
+  <dd>
+    Constructs a new {{URLPattern}} object.  The |options| is an object containing the additional configuration options that can affect how the components are matched.  Currently it has only one property {{URLPatternOptions/ignoreCase}} which can be set to true to have case-insensitive matching for pathname, search, and hash.
+
+    Note that by default, that is in the absence of the |options| argument, matching is always case-sensitive.
+  </dd>
+
+  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL, options)|URLPattern}}(|input|, |baseURL|, |options|)</code></dt>
```
wanderview, line 209: This case should use `|patternString|` instead of `|input|`.

spec.bs
```diff
@@ -184,16 +189,28 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
...
 
-  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL)|URLPattern}}(|patternString|, |baseURL|)</code></dt>
+  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL, options)|URLPattern}}(|patternString|, |baseURL|)</code></dt>
   <dd>
     Constructs a new {{URLPattern}} object.  |patternString| is a URL string containing pattern syntax for one or more components.  If |baseURL| is provided, then |patternString| can be relative.  This constructor will always set at least an empty string value and does not default any components to wildcard patterns.
   </dd>
 
+  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL, options)|URLPattern}}(|input|, |options|)</code></dt>
+  <dd>
+    Constructs a new {{URLPattern}} object.  The |options| is an object containing the additional configuration options that can affect how the components are matched.  Currently it has only one property {{URLPatternOptions/ignoreCase}} which can be set to true to have case-insensitive matching for pathname, search, and hash.
```
wanderview, line 204: Maybe reword the last sentence to:

"Currently it has only one property, {{URLPatternOptions/ignoreCase}}, which can be set to true to enable case-insensitive matching."

spec.bs
```diff
@@ -184,16 +189,28 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>search componen
...
+
+    Note that by default, that is in the absence of the |options| argument, matching is always case-sensitive.
+  </dd>
+
+  <dt><code>|urlPattern| = new {{URLPattern/constructor(input, baseURL, options)|URLPattern}}(|input|, |baseURL|, |options|)</code></dt>
+  <dd>
+    Constructs a new {{URLPattern}} object just like the previous overload, additionally supporting |baseURL|.
```
wanderview, line 211: Maybe change this entry to say something like:

"A {{URLPatternOptions}} object may also be passed when constructing a pattern from a string and base URL."

spec.bs
```diff
@@ -281,10 +301,13 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
-  1. If the result of running [=protocol component matches a special scheme=] given [=this=]'s [=URLPattern/protocol component=] is true, then set [=this=]'s [=URLPattern/pathname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/pathname}}"], [=canonicalize a pathname=], and [=pathname options=].
-  1. Else set [=this=]'s [=URLPattern/pathname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/pathname}}"], [=canonicalize an opaque pathname=], and [=default options=]
-  1. Set [=this=]'s [=URLPattern/search component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/search}}"], [=canonicalize a search=], and [=default options=].
-  1. Set [=this=]'s [=URLPattern/hash component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/hash}}"], [=canonicalize a hash=], and [=default options=].
+  1. Let |defaultOptions| be the result of running [$Object.assign$]({}, [=default options=], {sensitive: !ignoreCase})
```
wanderview, line 304: Maybe name this |compileOptions| instead of |defaultOptions|.  It may not be default any more if we modify it for the ignore case attribute.

wanderview, line 304: Please add an `ignoreCase` attribute to the internal options definition here:

https://github.com/WICG/urlpattern/blob/main/spec.bs#L1075-L1081

I think keeping `ignoreCase` instead of switching to `sensitive` would be easier as we would not need to flip the logic.

wanderview, line 304: I'm unsure about using Object.assign here in spec language.  @domenic?

wanderview (changes requested):
Thanks for writing this!  Overall it looks reasonable, but I have a few suggestions and requests.  Also, I wouldn't mind having @domenic look at it for the object.assign usage.

spec.bs
domenic: Yes, you cannot use Object.assign in spec language. Just update the field directly.

Sayan751:
@wanderview @domenic Thanks for the review. I have made the changes, outlined. I think it should now be ready for the next round of review or merge.

spec.bs
```diff
@@ -273,6 +289,9 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
     1. [=Assert=]: |input| is a {{URLPatternInit}}.
     1. If |baseURL| is given, then throw a {{TypeError}}.
     1. Set |init| to |input|.
+  1. Let |ignoreCase| be false.
```
wanderview, line 292: It seems like we don't really need a separate `|ignoreCase|` variable here.  We can just reference `|options|[{{URLpatternOptions/ignoreCase}}]` below.  If we set the default value to false in the webidl, then we don't have to check for existence before referencing it.

spec.bs
```diff
@@ -148,6 +148,10 @@ dictionary URLPatternInit {
   USVString baseURL;
 };
 
+dictionary URLPatternOptions {
+  boolean ignoreCase;
```
wanderview, line 152: I think we should make this `= false` here to provide a default.

spec.bs
```diff
@@ -281,10 +300,13 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
-  1. Else set [=this=]'s [=URLPattern/pathname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/pathname}}"], [=canonicalize an opaque pathname=], and [=default options=]
-  1. Set [=this=]'s [=URLPattern/search component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/search}}"], [=canonicalize a search=], and [=default options=].
-  1. Set [=this=]'s [=URLPattern/hash component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/hash}}"], [=canonicalize a hash=], and [=default options=].
+  1. Let |compileOptions| be the [=default options=] with the ignoreCase property set to |ignoreCase|.
+  1. If the result of running [=protocol component matches a special scheme=] given [=this=]'s [=URLPattern/protocol component=] is true, then:
+    1. Let |pathCompileOptions| be the [=pathname options=] with the ignoreCase property set to |ignoreCase|.
```
wanderview, line 305: Similar rewording here.

spec.bs
```diff
@@ -1074,7 +1099,7 @@ A [=part=] has an associated <dfn for=part>suffix</dfn>, a string, initially the
...
+An <dfn>options</dfn> [=struct=] contains different settings that control how [=/pattern string=] behaves.  These options originally come from [path-to-regexp](https://github.com/pillarjs/path-to-regexp).  We only include the options that are modified within the URLPattern specification and exclude the other options.  For the purposes of comparison, this specification acts like [path-to-regexp](https://github.com/pillarjs/path-to-regexp) where `ignoreCase` (the original [path-to-regexp](https://github.com/pillarjs/path-to-regexp) options is `sensitive` that is synonymous to negated `ignoreCase`), `strict`, `start`, and `end`.  All the aforementioned properties are always set to true except `ignoreCase` which is set to false.
```
wanderview, line 1102: You need another paragraph below that includes a `<dfn for=options>ignore case</dfn>` to define the attribute.  Then you can reference it like `[=options/ignore case=]`.  And I think using the separate words instead of camel case to match style would be slightly preferred.

The second paragraph also explicitly needs to say `[=options/ignore case=] default to false` or we need to update the various definitions of `[=default options=]`, etc, to set the value to false.

spec.bs
```diff
@@ -365,7 +387,10 @@ A [=component=] has an associated <dfn for=component>group name list</dfn>, a [=
   1. If |input| is null, then set |input| to "`*`".
   1. Let |part list| be the result of running [=parse a pattern string=] given |input|, |options|, and |encoding callback|.
   1. Let (|regular expression string|, |name list|) be the result of running [=generate a regular expression and name list=] given |part list| and |options|.
-  1. Let |regular expression| be [$RegExpCreate$](|regular expression string|, "`u`").  If this throws an exception, catch it, and throw a {{TypeError}}.
+  1. Let |flags| be an empty string.
+  1. If |options|["ignoreCase"] is true then set |flags| to "`ui`".
```
wanderview, line 391: Since `|options|` is an internal struct the prefered syntax is a bit different than the webidl object.  Instead write it as `|options|'s [=options/ignore case=]`.

spec.bs
```diff
@@ -281,10 +300,13 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
...
-  1. If the result of running [=protocol component matches a special scheme=] given [=this=]'s [=URLPattern/protocol component=] is true, then set [=this=]'s [=URLPattern/pathname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/pathname}}"], [=canonicalize a pathname=], and [=pathname options=].
-  1. Else set [=this=]'s [=URLPattern/pathname component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/pathname}}"], [=canonicalize an opaque pathname=], and [=default options=]
-  1. Set [=this=]'s [=URLPattern/search component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/search}}"], [=canonicalize a search=], and [=default options=].
-  1. Set [=this=]'s [=URLPattern/hash component=] to the result of [=compiling a component=] given |processedInit|["{{URLPatternInit/hash}}"], [=canonicalize a hash=], and [=default options=].
+  1. Let |compileOptions| be the [=default options=] with the ignoreCase property set to |ignoreCase|.
```
wanderview, line 303: I think this might be slightly better written as:

```
1. Let |compileOptions| be a copy of [=default options=] with [=options/ignore case=] set to |options|[{{URLPatternOptions/ignoreCase}}].
```

wanderview (changes requested):
Thanks again.  I found a few more issues in the pass.

Note, there will probably be some delay in landing this PR even when its ready.  I'm unsure about landing this without starting the process to change chromium.  If you want to help there you could file a bug at crbug.com and include a link to to it in the commit message for this PR.

If you are interested in helping with changes to chromium I can support you with that.  Otherwise I will need to carve out some time to do that in the next few weeks.

spec.bs
Sayan751: Did you actually mean to add the default in the ctor, instead? I am asking, because I am not sure with the syntax. Thinking from TS perspective, it seems that the `dictionary` gets translated to `interface` and there such default values might be odd. Can you please confirm?

spec.bs
Sayan751: Thanks! I was also not sure how to put it. In my head, 'copy' was a bit ambiguous term. I considered 'clone' as well. Didn't sit right, and decided to avoid both 😄 However, I can go with 'copy', if I am overthinking about the ambiguity, and that's a non-issue.

spec.bs
Sayan751: Never mind. Looked up the Web IDL specs 😄

Sayan751:
@wanderview FYI created this to track the implementation task in Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1345036

wanderview (approved):
LGTM.  Thanks!  Note I will probably wait to merge until implementation is close to landing.

spec.bs
```diff
@@ -120,7 +120,7 @@ typedef (USVString or URLPatternInit) URLPatternInput;
 
 [Exposed=(Window,Worker)]
 interface URLPattern {
-  constructor(optional URLPatternInput input = {}, optional USVString baseURL);
+  constructor(optional URLPatternInput input = {}, optional USVString baseURL, optional URLPatternOptions options = {});
```
wanderview, line 123: I think we also need a second constructor form to do what you want:

```
 constructor(optional URLPatternInput input = {}, optional URLPatternOptions options = {});
```

Otherwise you cannot specify options without also specifying a base URL second arg.

wanderview: I'm going to fix this in a follow-up commit.

PR #169, opened 2022-09-27
Additional webidl fixes for constructor options. (#148)

wanderview:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/169.html" title="Last updated on Sep 28, 2022, 2:27 PM UTC (aebd51b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/169/7a09f4e...aebd51b.html" title="Last updated on Sep 28, 2022, 2:27 PM UTC (aebd51b)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -120,7 +120,8 @@ typedef (USVString or URLPatternInit) URLPatternInput;
 
 [Exposed=(Window,Worker)]
 interface URLPattern {
-  constructor(optional URLPatternInput input = {}, optional USVString baseURL, optional URLPatternOptions options = {});
+  constructor(URLPatternInput input, USVString baseURL, optional URLPatternOptions options = {});
+  constructor(optional URLPatternInput input = {}, optional URLPatternOptions options = {});
 
   boolean test(optional URLPatternInput input = {}, optional USVString baseURL);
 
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic PTAL

spec.bs
```diff
@@ -419,7 +432,7 @@ A [=component=] has an associated <dfn for=component>group name list</dfn>, a [=
     1. Set |hash| to |applyResult|["{{URLPatternInit/hash}}"].
   1. Else:
     1. Let |baseURL| be null.
-    1. If |baseURLString| was given, then:
+    1. If |baseURLString| was given and not null, then:
```
domenic, line 435: Probably better is to change the declaration to "an optional string-or-null |baseURLString| (default null)" and then change this to "is not null".

spec.bs
```diff
@@ -277,33 +278,45 @@ Each {{URLPattern}} object has an associated <dfn for=URLPattern>hash component<
   </dd>
 </dl>
 
+<div algorithm>
```
domenic, line 281: Their declaration order is opposite their definition order; it'd be slightly nicer if they were matching. (I clicked on the first one in the IDL block and was surprised I had to scroll up, instead of down, to see the second one.)

domenic (approved):
LGTM with nits

spec.bs
wanderview: While looking at this I realized this baseURLString is not the same as the constructor baseURL.  I will just revert this change as I think its better just being strictly optional.

PR #172, opened 2022-11-11
Escape pattern chars in pattern base URLs. (Fixes #170)

### LINKED ISSUE #170: Base URL handling should escape any pattern characters
Currently, if the base URL contains any special pattern characters such as `+` and `*`, these gain special meaning in the pattern. Instead, they should have whatever escaping is required to make them simply literal characters.

Currently, things such as this fail: `new URLPattern({baseURL: 'https://chromium-review.googlesource.com/c/chromium/src/+/123456'})`. Other URLs might be valid but introduce unintended wildcards.

I think it's reasonable for authors to expect that `new URLPattern({baseURL})` will succeed if `baseURL` is a valid URL, and that it will match exactly that URL.

wanderview:
***
<a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/pull/172.html" title="Last updated on Nov 16, 2022, 3:44 PM UTC (49a6522)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/urlpattern/172/f6881dd...49a6522.html" title="Last updated on Nov 16, 2022, 3:44 PM UTC (49a6522)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1721,14 +1721,14 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   1. If |init|["{{URLPatternInit/baseURL}}"] is not null:
     1. Set |baseURL| to the result of [=URL parser|parsing=] |init|["{{URLPatternInit/baseURL}}"].
     1. If |baseURL| is failure, then throw a {{TypeError}}.
-    1. Set |result|["{{URLPatternInit/protocol}}"] to |baseURL|'s [=url/scheme=].
-    1. Set |result|["{{URLPatternInit/username}}"] to |baseURL|'s [=url/username=].
-    1. Set |result|["{{URLPatternInit/password}}"] to |baseURL|'s [=url/password=].
-    1. Set |result|["{{URLPatternInit/hostname}}"] to |baseURL|'s [=url/host=] or the empty string if the value is null.
-    1. Set |result|["{{URLPatternInit/port}}"] to |baseURL|'s [=url/port=] or the empty string if the value is null.
-    1. Set |result|["{{URLPatternInit/pathname}}"] to the result of [=URL path serializing=] |baseURL|.
-    1. Set |result|["{{URLPatternInit/search}}"] to |baseURL|'s [=url/query=] or the empty string if the value is null.
-    1. Set |result|["{{URLPatternInit/hash}}"] to |baseURL|'s [=url/fragment=] or the empty string if the value is null.
+    1. Set |result|["{{URLPatternInit/protocol}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/scheme=] and |type|.
+    1. Set |result|["{{URLPatternInit/username}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/username=] and |type|.
+    1. Set |result|["{{URLPatternInit/password}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/password=] and |type|.
+    1. Set |result|["{{URLPatternInit/hostname}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/host=] and |type|.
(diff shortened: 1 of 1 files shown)


wanderview:
@domenic PTAL.  Thanks.

Corresponding chromium CL:

https://chromium-review.googlesource.com/c/chromium/src/+/4020802

spec.bs
```diff
@@ -1721,14 +1721,14 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
     1. If |baseURL| is failure, then throw a {{TypeError}}.
-    1. Set |result|["{{URLPatternInit/protocol}}"] to |baseURL|'s [=url/scheme=].
-    1. Set |result|["{{URLPatternInit/username}}"] to |baseURL|'s [=url/username=].
-    1. Set |result|["{{URLPatternInit/password}}"] to |baseURL|'s [=url/password=].
-    1. Set |result|["{{URLPatternInit/hostname}}"] to |baseURL|'s [=url/host=] or the empty string if the value is null.
-    1. Set |result|["{{URLPatternInit/port}}"] to |baseURL|'s [=url/port=] or the empty string if the value is null.
-    1. Set |result|["{{URLPatternInit/pathname}}"] to the result of [=URL path serializing=] |baseURL|.
-    1. Set |result|["{{URLPatternInit/search}}"] to |baseURL|'s [=url/query=] or the empty string if the value is null.
-    1. Set |result|["{{URLPatternInit/hash}}"] to |baseURL|'s [=url/fragment=] or the empty string if the value is null.
+    1. Set |result|["{{URLPatternInit/protocol}}"] to the result of running [=process a base URL string=] given |baseURL|'s [=url/scheme=] and |type|.
```
domenic, line 1724: Nit: with verbs like this you could say `the result of [=processing a base URL string=] given`

spec.bs
```diff
@@ -1721,14 +1721,14 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+    1. Set |result|["{{URLPatternInit/username}}"] to the result of running [=process a base URL string=] given |baseURL|'s [=url/username=] and |type|.
+    1. Set |result|["{{URLPatternInit/password}}"] to the result of running [=process a base URL string=] given |baseURL|'s [=url/password=] and |type|.
+    1. Set |result|["{{URLPatternInit/hostname}}"] to the result of running [=process a base URL string=] given |baseURL|'s [=url/host=] and |type| or the empty string if the value is null.
```
domenic, line 1727: I think these sentences got a bit confusing. It's no longer clear what "or the empty string" attaches to.

Maybe "process a base URL string" could return the empty string if its first argument is null?

spec.bs
```diff
@@ -1755,6 +1755,13 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   1. Return |result|.
 </div>
 
+<div algorithm>
+  To <dfn>process a base URL string</dfn> given a string |input| and a string |type|:
+
+  1. If |type| is not "`pattern`" return |input|.
+  1. Return the result of running [=escape a pattern string=] given |input|.
```
domenic, line 1762: Similarly, `[=escaping a pattern string=]`

spec.bs
wanderview: That is a little different since previously we did not always check for null, but I think its ok to do this for base URL.  Alternatively, we could assert the component is not null because I believe URL parsing will not produce a null component value.

Anyway, I moved the check to the algorithm for now.

wanderview: Actually, I moved it to an assert.

spec.bs
```diff
@@ -1755,6 +1755,14 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   1. Return |result|.
 </div>
 
+<div algorithm>
+  To <dfn>process a base URL string</dfn> given a string |input| and a string |type|:
+
+  1. [=Assert=] that |input| is not null.
```
domenic, line 1761:
```suggestion
  1. [=Assert=]: |input| is not null.
```

domenic (approved):

PR #184, opened 2023-09-12
Fix illegal closing tags.

jeremyroman:


sisidovski (approved):

domenic:
Nit on the commit message: if you want to follow https://github.com/whatwg/meta/blob/main/COMMITTING.md#commit-messages , no commit message. And probably a "Meta:" prefix since it is related to just getting the spec to build and isn't noticeable at all (even as an editorial change) in the output.

jeremyroman:
> Nit on the commit message: if you want to follow https://github.com/whatwg/meta/blob/main/COMMITTING.md#commit-messages , no commit message. And probably a "Meta:" prefix since it is related to just getting the spec to build and isn't noticeable at all (even as an editorial change) in the output.

Done.

Issue #190, opened 2023-10-05
WHATWG migration


sisidovski:
As we discussed in https://github.com/whatwg/sg/issues/215, we're going to move URLPattern into WHATWG.

Here is the checklist that we should cover.

- [x] Reach WHATWG SG agreement: https://github.com/whatwg/sg/issues/215
- [x] Transfer repository from WICG to WHATWG
- [x] Create a logo: https://github.com/whatwg/whatwg.org/pull/427 and generally complete the checklist in https://github.com/whatwg/meta/blob/main/NEW-STANDARD.md
- [x] Update the document in this repository to account for this move
- [x] Add redirect for github.io URLs ([like this](https://github.com/WICG/wicg.github.io/blob/main/picture-in-picture/index.html)): https://github.com/WICG/wicg.github.io/pull/22

cc @domenic @jeremyroman

domenic:
(Note that I edited the last item on the checklist to reflect how we _can_ redirect the github.io URLs.)

domenic:
Given that we've broken https://wicg.github.io/urlpattern/, would one of the editors be willing to send a PR to add a redirect to the https://github.com/WICG/wicg.github.io repository?

Note that https://urlpattern.spec.whatwg.org/ doesn't display anything quite yet, but redirecting to that page is probably just as good as 404ing, and it's something we can do ahead of time.

sisidovski:
I'm happy to take that task.

sisidovski:
Sent https://github.com/WICG/wicg.github.io/pull/22

domenic:
All done!

Some potential followup work in https://github.com/whatwg/urlpattern/issues/195.

PR #193, opened 2023-10-17
Editorial: update the standard for the WHATWG move

domenic:
* Update repository files using whatwg/spec-factory.
* Use the WHATWG template.
* Add an Acknowledgments section and IPR statement.
* Sentence-case all headings.
* Add missing example IDs and suppress a warning about "optional" being used in an example.
* Minor fixes to the introductory example to work well with WHATWG styling.
* Fix some invalid HTML, double-"the" typos, and usage of "and/or" disallowed by the linter.
* Reword some disallowed uses of RFC 2119 keywords.

Part of #190. See also https://github.com/whatwg/sg/issues/215.

I'd like editor review and also @annevk review if possible.

***

### :boom: Error: 400 Bad Request :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Oct 18, 2023, 5:54 AM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fwhatwg%2Furlpattern%2F13b8b1464ec873dab131332035707adba66a8bb0%2Fspec.bs&md-warning=not%20ready)

```
Error running preprocessor, returned code: 2.
FATAL ERROR: Found unmatched text macro [COMMIT-SHA]. Correct the macro, or escape it with a leading backslash.
 ✘  Did not generate, due to errors exceeding the allowed error level.
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20whatwg/urlpattern%23193.)._
</details>


spec.bs
```diff
@@ -1,21 +1,10 @@
...
-Boilerplate: omit conformance, omit feedback-header
-Editor: Ben Kelly, Google https://www.google.com/, wanderview@chromium.org
-Abstract: The URLPattern API provides a web platform primitive for matching URLs based on a convenient pattern syntax.
-Status Text: This proposal will hopefully move into the WHATWG URL workstream in the future.
-!Participate: <a href="https://github.com/WICG/urlpattern">GitHub WICG/urlpattern</a> (<a href="https://github.com/WICG/urlpattern/issues/new">new issue</a>, <a href="https://github.com/WICG/urlpattern/issues?state=open">open issues</a>)
-!Commits: <a href="https://github.com/WICG/urlpattern/commits/main/spec.bs">GitHub spec.bs commits</a>
-Complain About: accidental-2119 yes, missing-example-ids yes
+Text Macro: TWITTER urlpatternapi
+Abstract: The URL Pattern Standard provides a web platform primitive for matching URLs based on a convenient pattern syntax.
 Indent: 2
```
domenic, line 7: We'll update spec_factory to allow 2-space indents here instead of the usual 1-space.

domenic:
The build will fail until we run spec-factory to update the Makefile, etc.; that will be a separate pull request.

spec.bs
```diff
@@ -1873,3 +1804,21 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+Luca Casonato, and
+Sayan Pal
+for their contributors to this specification.
+
+A special thanks to the [pillarjs/path-to-regexp](https://github.com/pillarjs/path-to-regexp) creators for their work on the open-source library and pattern syntax which inspired this standard.
+
+This standard is written by
+Ben Kelly ([Google](https://www.google.com/), [wanderview@chromium.org](mailto:wanderview@chromium.org)),
+Jeremy Roman ([Google](https://www.google.com/), [jbroman@chromium.org](mailto:jbroman@chromium.org)), and
+Shunya Shishido ([Google](https://www.google.com/), [sisidovski@chromium.org](mailto:sisidovski@chromium.org)).
```
domenic, line 1821: These should stay synced with https://github.com/whatwg/sg/pull/218/files ; if either changes let's be sure to update the other PR.

annevk:
Setting up the tooling would be good because then PR Preview works and we can see what the document ends up looking like, right?

domenic:
Yeah, I'll try to land that first, but I think that depends on db.json.

jeremyroman:
Looks fine on inspection, but a preview would be nice. :)

domenic:
I think a preview is not possible because PR preview is using the old config file against the new spec, or something like that, and thus fails the whole process when it can't generate the old spec in order to create diffs.

In lieu of that I've set up a preview here: https://boom-bath.glitch.me/urlpattern-preview.html

Editor review especially appreciated for changes to spec.bs.

annevk (approved):

spec.bs
```diff
@@ -426,7 +357,7 @@ A [=component=] has an associated <dfn for=component>group name list</dfn>, a [=
   1. If |options|'s [=options/ignore case=] is true then set |flags| to "`vi`".
   1. Otherwise set |flags| to "`v`"
   1. Let |regular expression| be [$RegExpCreate$](|regular expression string|, |flags|).  If this throws an exception, catch it, and throw a {{TypeError}}.
-    <p class="note allow-2119">The specification uses regular expressions to perform all matching, but this is not required.  Implementations are free to perform matching directly against the [=/part list=] when possible; e.g. when there are no custom regexp matching groups.  If there are custom regular expressions, however, its important that they should be immediately evaluated in the [=compile a component=] algorithm so an error can be thrown if they are invalid.
```
sisidovski, line 429: `should` seems to be just removed. Don't we need an alternative word?

sisidovski (approved):

spec.bs
domenic: Thank you for the careful review! I think it is OK in this case actually. "It is important that they be (verb phrase)" is acceptable.

domenic:
The standard is live! https://urlpattern.spec.whatwg.org/

PR #194, opened 2023-10-17
Fix property descriptions in the quick reference

zcorpan:


jeremyroman:
Looks reasonable to me (well, we could bikeshed scheme/protocol, path/pathname and hash/fragment, but it probably doesn't matter in this context). Unfortunately I seem to not have been granted permission to review PRs since this was moved into whatwg.

@domenic do you know what's required for that?

sisidovski (approved):
LGTM unless we have bikeshedding.

domenic:
https://github.com/whatwg/urlpattern/issues/195 might be relevant here.

Yes, both of you have the same permissions; Jeremy please try again?

jeremyroman (approved):

jeremyroman:
I was able to approve, but I'm still not authorized to merge PRs due to branch protection.

domenic:
I found a setting that seemed different between this and other standards. Give it a shot now?

PR #198, opened 2023-10-27
Use "inherit left, wildcard right" behavior in base URL application and constructor string parsing.

### LINKED ISSUE #179: Base URL inheritance gives unintuitive results
(For background see https://github.com/WICG/nav-speculation/issues/259; in particular this is attempting to formalize the proposal at https://github.com/WICG/nav-speculation/issues/259#issuecomment-1507870169. cc @jeremyroman)

### Problem

When constructing a `URLPattern` from a base URL, we have found it surprising how many components get inherited. The particularly problematic ones are `search` and `hash`. Consider code such as:

```js
const pattern1 = new URLPattern("/wp-login.php", document.baseURI);

// or:

const pattern2 = new URLPattern({ pathname: "/wp-login.php", baseURL: document.baseURI });

if (!pattern1.test(userInput)) {
  // They're not targeting the login page for our blog. OK to proceed!
}
```

Can you spot the bug? There are actually two!

- If `document.baseURL` is something like `https://example.com/`, then this sort of check gets defeated by a `userInput` derived from a relative URL such as `/wp-login.php?bypass` or `/wp-login.php#bypass`.
- If `document.baseURL` is something like `https://example.com/?utm_source=blah`, then this sort of check fails even for `userInput` derived from `/wp-login.php`.

This is because these `URLPattern` instances inherited the empty-string `search` and `hash` components from the base URL:

```js
const noBaseURL = new URLPattern("/wp-login.php");
console.assert(noBaseURL.search === "*");
console.assert(noBaseURL.hash === "*");

const withBaseURL1 = new URLPattern("/wp-login.php", "https://example.com/");
console.assert(withBaseURL1.search === "");
console.assert(withBaseURL1.hash === "");

const withBaseURL2 = new URLPattern("/wp-login.php", "https://example.com/?utm_source=blah");
console.assert(withBaseURL2.search === "utm_source=blah");
console.assert(withBaseURL2.hash === "");
```

#### Non-path/search/hash cases

Another instance that some *might* find unexpected is the following:

```js
const pattern = new URLPattern({ protocol: "https", baseURL: document.baseURI });
```

This pattern inherits all non-`protocol` components from the base URL. Whereas, some might expect that since we overrode such an "early" component, we'd only get matching on `protocol`, not the rest.

This is especially problematic in environments where the base URL argument is implicit, e.g. [speculation rules](https://github.com/WICG/nav-speculation/blob/main/triggers.md#document-rules), [service worker static routing](https://github.com/yoshisatoyanagisawa/service-worker-static-routing-api), or any framework which attempts to specialize itself to "the current page", such as

```js
function addRouteHandlerForThisPage(urlPatternString, handler) {
  routerTable.set(new URLPattern(urlPatternString, document.baseURI), handler);
}
```

In such environments, you might provide a URL pattern such as

```json
{
  "href_matches": { "protocol": "https" }
}
```

or

```js
condition: {
  urlPattern: { protocol: "https" }
}
```

and expect this to match all `https://` URLs. But instead, since it inherits all non-protocol components from the base URL, it essentially only matches the base URL itself.

### Proposed solution

- Whenever a `URLPattern` input has a given component, do not (by default) inherit any components from the base URL that are "later" than that component. "Later" means mostly the usual order: `protocol`, `hostname`, `port`, `pathname`, `search`, `hash`. For example:
  - If the input has a `protocol` component, do not inherit any components
  - If the input has a `pathname` component, do not inherit the `search` and `hash` components
- For `username` and `password`, we define those to be "later" than `protocol`, `hostname`, and `port` (and define `password` to be "later" than `username`). For example:
  - If the input has a `hostname` component, do not inherit `port`, `pathname`, `search`, `hash`, `username`, and `password`
  - If the input has a `username` component, do not inherit `pathname`, `search`, `hash`, `password`
- Add an option, `baseURLInheritance`, which lets you control this behavior
  - The new default, per the above, is `"auto"`
  - `"all"` gives the current behavior
  - Maybe, you could also do an array of components to inherit, e.g. `["port", "search"]`. We can leave that out for now minus compelling use cases.

Then:

```js
const pattern1 = new URLPattern("/wp-login.php", "https://example.com/");
console.assert(pattern1.protocol === "https");
console.assert(pattern1.username === "");
console.assert(pattern1.password === "");
console.assert(pattern1.hostname === "example.com");
console.assert(pattern1.port === "");
console.assert(pattern1.pathname === "/wp-login.php");
console.assert(pattern1.search === "*");
console.assert(pattern1.hash === "*");

const pattern2 = new URLPattern("/wp-login.php", "https://example.com/?utm_source=blah");
// Same as pattern1, including:
console.assert(pattern2.search === "*");
console.assert(pattern2.hash === "*");

const pattern3 = new URLPattern("/wp-login.php", "https://example.com/?utm_source=blah#heading");
// Same as pattern1, including:
console.assert(pattern3.search === "*");
console.assert(pattern3.hash === "*");


const pattern4 = new URLPattern("/wp-login.php?user=foo", "https://example.com/");
// Same as pattern1 before search. Then:
console.assert(pattern4.search === "user=foo");
console.assert(pattern4.hash === "*");

const pattern5 = new URLPattern("/wp-login.php?user=foo#bar", "https://example.com/");
// Same as pattern1 before search. Then:
console.assert(pattern5.search === "user=foo");
console.assert(pattern5.hash === "bar");


const pattern6 = new URLPattern("/wp-login.php", "https://example.com/?utm_source=blah#heading", { baseURLInheritance: "all" });
console.assert(pattern6.search === "utm_source=blah");
console.assert(pattern6.hash === "heading");
```

#### Compat

There might be some compat impact here. I suspect it is low, as we'd need to:

- Be using `URLPattern`
- ...with a base URL containing a given component
- ...and a pattern that does not contain that component
- ...executed against an input where it makes a difference

"using `URLPattern`" is [swinging between 0.02% and 0.05% of page views](https://chromestatus.com/metrics/feature/timeline/popularity/3959), so we're in a good place to start with. We plan to add use counters to see how often this conjunction happens in the wild.

#### Alternatives considered

- We could make this new behavior opt-in instead of opt-out. I would strongly prefer to make it opt-out, because: I think it's more sensible and less likely to lead to bugs; and, I think it's the best default for various web platform features, such as speculation rules or service worker scope matching, which hope to base themselves on URL patterns.

- We could just tell people to always write their patterns like `"/wp-login.php?*#*`. I think that would be a sad place to end up.

- We could restrict this logic to just `pathname` -> `search` -> `hash`. I think this is reasonable as those are the most problematic components. This would mean not solving the `{ protocol: "https" }` problem, but maybe that's OK.

#### Server-side considerations

From my experience, most server-side uses of `URLPattern` are pathname-focused. They probably aren't overly impacted by this. Any confirmation from the server-side community would be helpful.

#### Who does the work?

Jeremy and I are happy to volunteer on the implementation/spec/test work. Jeremy is working on the use counter.

jeremyroman:
The following changes apply to patterns which are constructed using a base URL, the constructor string syntax, or both -- but not any pattern which explicitly specifies components separately without a base URL.

* Components are not inherited from a base URL if an "earlier" component is explicitly specified.
* In the string format, unspecified "later" components are implicitly wildcarded, rather than required to be empty (with the exception of the port, which is always taken to be specified when the hostname is).
* Username and password are never implicitly specified or inherited.

For example:
1. `"https://example.com/*"` also matches with any username, password, search, and hash. Previously this would be written `"https://*:*@example.com/*\\?*#*"`.
2. `new URLPattern({ pathname: "/login" }, "https://example.com/?q=hello")` accepts any query string and hash, not only `"?q=hello"` and `""`.
3. `"https://*:*"` or `{protocol: "https"}` means "any HTTPS URL, on any port", and `"https://*"` means "any HTTPS URL on the default port (443)". These have the same meaning whether or not a base URL is provided, since specifying the protocol prohibits inheritance of other components.

This makes patterns more expansive than before, in cases where wildcards are likely to be desirable.

The logic of inheriting components from a base URL dictionary is also similarly changed in a way that may make it _not_ match where it did before, but more consistently with the above and with how relative URL strings are resolved. For example, `new URLPattern("https://example.com/foo?q=1#hello").test({pathname: "/foo", hash: "hello", baseURL: "https://example.com/bar?q=1"})` previously returned `true` but will now be `false`, since the search component is not inherited when the pathname is specified. This is analogous to how `new URL("/foo#hello", "https://example.com/bar?q=1")` works. The reverse is also possible; in both cases this is quite niche.

Fixes #179.

---

   * Google Chrome
   * Safari (per https://github.com/whatwg/urlpattern/issues/179#issuecomment-1735850741)
   * https://chromium-review.googlesource.com/c/chromium/src/+/4975595
   * Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1468446
   * Deno: https://github.com/denoland/deno/issues/21073
   * kenchris/urlpattern-polyfill: https://github.com/kenchris/urlpattern-polyfill/issues/122
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/content/issues/30024

***
<a href="https://whatpr.org/urlpattern/198.html" title="Last updated on Nov 7, 2023, 6:15 PM UTC (7cad49e)">Preview</a> | <a href="https://whatpr.org/urlpattern/198/c8d5072...7cad49e.html" title="Last updated on Nov 7, 2023, 6:15 PM UTC (7cad49e)">Diff</a>


jeremyroman:
@sisidovski @domenic FYI

jeremyroman:
Updated commit message in the opening comment (hopefully it's reasonably clear).

I assume Gecko and WebKit implementation bugs are not desired since they don't yet (to my knowledge) implement URLPattern; is that right? I can file Deno and urlpattern-polyfill ones shortly before merging if appropriate. Ditto for MDN.

For implementer interest, it sounded like Anne was on board with this direction; what level of assent is required to add WebKit to the implementer interest list?

domenic:
> Updated commit message in the opening comment (hopefully it's reasonably clear).

LGTM! You might want to add something about the compatibility implications, and how we've measured them to be low.

> I assume Gecko and WebKit implementation bugs are not desired since they don't yet (to my knowledge) implement URLPattern; is that right?

Right, we can treat them as being under the umbrella "implement URLPattern" bugs.

> For implementer interest, it sounded like Anne was on board with this direction; what level of assent is required to add WebKit to the implementer interest list?

What I would do is check the box for WebKit, linking to https://github.com/whatwg/urlpattern/issues/179#issuecomment-1735850741 , but also tag @annevk to confirm, and give a few days before merging in case he has objections or wants to take a closer look.

jeremyroman:
> What I would do is check the box for WebKit, linking to [#179 (comment)](https://github.com/whatwg/urlpattern/issues/179#issuecomment-1735850741) , but also tag @annevk to confirm, and give a few days before merging in case he has objections or wants to take a closer look.

Alright, will do unless @annevk objects.

spec.bs
```diff
@@ -510,16 +510,16 @@ A [=constructor string parser=] has an associated <dfn export for="constructor s
...
-      <p>To implement this we initialize components in |parser|'s [=constructor string parser/result=] with empty string in advance.
-      <p>We can't, however, do this immediately.  We want to allow the `baseURL` to provide information for relative URLs, so we only want to set the default empty string values for components following the first component in the relative URL.  We therefore wait to set the default component values until after we exit the "<a for="constructor string parser/state">`init`</a>" [=constructor string parser/state=].
+      <p>When constructing a pattern using a {{URLPatternInit}} like `new URLPattern({ pathname: 'foo' })` any missing components will be defaulted to wildcards.
+      <p>Components which appear "later" than those specified are instead treated as wildcards (see above). As a result, these values are not initialized to be empty in |parser|'s [=constructor string parser/result=] until a "later" component is seen.
```
domenic, line 522: The new version of this note doesn't make a lot of sense. Previously, it was contrasting `new URLPattern({ pathname: 'foo' })` and the constructor string parser. Now, it's stating two disconnected facts: one about the `{ pathname: 'foo' }` version, and one about the constructor string parser. And for some reason it's using "instead" to contrast them, even though the two disconnected facts are actually very similar.

I think maybe this note could be deleted, now that the treatment is more symmetric? Or maybe something like the following:

> When constructing a pattern using a {{URLPatternInit}} like `new URLPattern({ pathname: 'foo' })` any missing components will be defaulted to wildcards. For the constructor string case, a somewhat-similar behavior applies: missing components are treated as either the empty string, or as wildcards, depending on what components _do_ appear in the input string. In particular, missing components which appear "later" than present components are treated as wildcards, whereas missing components that appear "earlier" than present components are treated as the empty string.
>
> To accomplish this behavior, missing components aren't set in the parser's result until a later component is encountered, as part of the change state algorithm.

spec.bs
```diff
@@ -642,13 +630,22 @@ To <dfn>parse a constructor string</dfn> given a string |input|:
...
+    1. If |parser|'s [=constructor string parser/state=] is "<a for="constructor string parser/state">`protocol`</a>", "<a for="constructor string parser/state">`authority`</a>", "<a for="constructor string parser/state">`username`</a>", "<a for="constructor string parser/state">`password`</a>", "<a for="constructor string parser/state">`hostname`</a>", or "<a for="constructor string parser/state">`port`</a>"; |new state| is "<a for="constructor string parser/state">`search`</a>", or "<a for="constructor string parser/state">`hash`</a>"; and |parser|'s [=constructor string parser/result=]["{{URLPatternInit/pathname}}"] does not [=map/exist=], then:
```
domenic, line 645:
```suggestion
    1. If |parser|'s [=constructor string parser/state=] is "<a for="constructor string parser/state">`protocol`</a>", "<a for="constructor string parser/state">`authority`</a>", "<a for="constructor string parser/state">`username`</a>", "<a for="constructor string parser/state">`password`</a>", "<a for="constructor string parser/state">`hostname`</a>", or "<a for="constructor string parser/state">`port`</a>"; |new state| is "<a for="constructor string parser/state">`search`</a>" or "<a for="constructor string parser/state">`hash`</a>"; and |parser|'s [=constructor string parser/result=]["{{URLPatternInit/pathname}}"] does not [=map/exist=], then:
```

spec.bs
```diff
@@ -1686,16 +1683,26 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   1. Set |result|["{{URLPatternInit/hash}}"] to |hash|.
   1. Let |baseURL| be null.
   1. If |init|["{{URLPatternInit/baseURL}}"] is not null:
+    <div class="note">
+      The base URL can be used to supply additional context, but for each component, if |init| includes a component which is at least as specific as one in the base URL, none is inherited.
+
+      A component is more specific if it appears later in one of the following two lists (which are very similar to the order they appear in the URL syntax):
```
domenic, line 1689: Is this notion of "more specific" shared by both inheritance schemes? If so, maybe this helpful note should be moved somewhere more generic? Then the discussions of "later" can link to it.

However, they do seem subtlely different, so probably there's not much to be done here. (I guess, is it possible to give a similar nice list-based explanation for the constructor string parsing?)

spec.bs
```diff
@@ -1686,16 +1683,26 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+
+      * protocol, hostname, port, pathname, search, hash
+      * protocol, hostname, port, username, password
+
+      Username and password are also never inherited when constructing patterns, only for a URL being matched.
```
domenic, line 1694:
```suggestion
      Username and password are also never inherited from the base URL when constructing a new {{URLPattern}} from a {{URLPatternInit}}. (They are, however, inherited from the base URL supplied as an argument to {{URLPattern/test()}} or {{URLPattern/exec()}}, for the URL being matched.)
```

spec.bs
```diff
@@ -1686,16 +1683,26 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   1. Set |result|["{{URLPatternInit/hash}}"] to |hash|.
   1. Let |baseURL| be null.
   1. If |init|["{{URLPatternInit/baseURL}}"] is not null:
```
domenic, line 1685: I think this algorithm has a persistent, existing problem of conflating "not present" and null. This is worse now because you're doing the right thing and doing "exists" checks in your new spec text, but the spec text right below it is doing "null" checks.

Probably you should fix them all to be "exists" checks.

spec.bs
```diff
@@ -1686,16 +1683,26 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
-    1. Set |result|["{{URLPatternInit/port}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/port=] and |type|.
-    1. Set |result|["{{URLPatternInit/pathname}}"] to the result of [=processing a base URL string=] given the result of [=URL path serializing=] |baseURL| and |type|.
-    1. Set |result|["{{URLPatternInit/search}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/query=] and |type|.
-    1. Set |result|["{{URLPatternInit/hash}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/fragment=] and |type|.
+    1. If |init|["{{URLPatternInit/protocol}}"] does not [=map/exist=], then set |result|["{{URLPatternInit/protocol}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/scheme=] and |type|.
+    1. If |type| is not "`pattern`" and |init| [=map/contains=] none of "{{URLPatternInit/protocol}}", "{{URLPatternInit/hostname}}", "{{URLPatternInit/port}}" and "{{URLPatternInit/username}}", then set |result|["{{URLPatternInit/username}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/username=] and |type|.
```
domenic, line 1699: and -> or here and below, I think?

domenic (approved):
Normatively LGTM, just editorial suggestions

annevk:
I didn't do an in-depth review, but happy to agree on behalf of WebKit with the behavior we're aiming for here.

spec.bs
jeremyroman: Done, with slightly different wording.

spec.bs
jeremyroman: I'm unsure. The WHATWG style guide is silent and I haven't found another style reference, probably because "none of" in formal writing generally doesn't go directly into a list. To me "none of A, B, and C" seems most natural (i.e., none of the elements in this set, which consists of A, B, and C), but "or" doesn't sound wrong either.

Another alternative would be "neither...nor", but for three or more items it's usual to repeat "nor" which gets pretty tedious for like 7 items that we get to (even more than the famous [USPS creed](https://en.wikipedia.org/wiki/United_States_Postal_Service_creed)):
> If init contains neither "protocol", nor "hostname", nor "port", nor "pathname", nor "search", nor "hash", ...

Yet another would be to use "any" instead of "none" and move the negation earlier, but it reads as a little more verbose too:
> If init does not contain any of "protocol", "hostname", "port", "pathname", "search", and "hash", ...

I don't think it's ambiguous in any case; what do you think?

spec.bs
```diff
@@ -71,6 +71,97 @@ It can be constructed using a string for each component, or from a shorthand str
...
+    <dd>"`discussion.example`"
+
+    <dt>[=URLPattern/port component|port=]
+    <dd>""
+
+    <dt>[=URLPattern/pathname component|pathname=]
+    <dd>"`/admin/*`"
+
+    <dt>[=URLPattern/search component|search=]
+    <dd>"*"
```
domenic, line 144:
```suggestion
    <dd>"`*`"
```

spec.bs
```diff
@@ -510,17 +601,14 @@ A [=constructor string parser=] has an associated <dfn export for="constructor s
...
   <p>Finally, the URLPattern constructor string parser does not handle some parts of the [=basic URL parser=] state machine.  For example, it does not treat backslashes specially as they would all be treated as pattern characters and would require excessive escaping.  In addition, this parser might not handle some more esoteric parts of the URL parsing algorithm like file URLs with a hostname.  The goal with this parser was to handle the most common URLs while allowing any niche case to be handled instead via the {{URLPatternInit}} constructor.
+  <p>In the constructor string algorithm, the pathname, search, and hash are wildcarded if earlier components are specified but later ones are not. For example, "`https://example.com/foo`" matches any search and any hash. Similarly, "`https://example.com`" matches any URL on that origin. This is analogous to the notion of a more specific component in the notes about [=process a URLPatternInit=] (e.g., a search is more specific than a pathname), but the constructor syntax only has a few cases where it is possible to specify a more specific component without also specifying the less specific components.
```
domenic, line 604: Maybe insert a `</div><div class="note">` before this since logically it feels like there are two notes here.

spec.bs
domenic: LGTM

spec.bs
```diff
@@ -1685,23 +1770,33 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+  1. If |init|["{{URLPatternInit/baseURL}}"] [=map/exists=]:
+    <div class="note">
+      The base URL can be used to supply additional context, but for each component, if |init| includes a component which is at least as specific as one in the base URL, none is inherited.
+
+      A component is more specific if it appears later in one of the following two lists (which are very similar to the order they appear in the URL syntax):
+
+      * protocol, hostname, port, pathname, search, hash
+      * protocol, hostname, port, username, password
+
+      Username and password are also never inherited from a base URL when constructing a {{URLPattern}}. (They are, however, inherited from the base URL supplied as an argument to {{URLPattern/test()}} or {{URLPattern/exec()}}.)
```
domenic, line 1782:
```suggestion
      Username and password are also never inherited from a base URL when constructing a {{URLPattern}}. (They are, however, inherited from the base URL when parsing a URL from the string supplied as an argument to {{URLPattern/test()}} or {{URLPattern/exec()}}.)
```

spec.bs
domenic: ChatGPT agrees with "and": https://chat.openai.com/share/550e4e1f-39cf-4a01-8664-85cdbb399d2b. (Bard is incoherent.)

I might go with

> If init does not contain "protocol", "hostname", "port", "pathname", "search", or "hash", ...

but I don't think we need to spend any more time on this. I agree people will understand either way :)

spec.bs
```diff
@@ -1685,23 +1770,33 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+    1. If |init| [=map/contains=] none of "{{URLPatternInit/protocol}}", "{{URLPatternInit/hostname}}", "{{URLPatternInit/port}}", "{{URLPatternInit/pathname}}", and "{{URLPatternInit/search}}", then set |result|["{{URLPatternInit/search}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/query=] and |type|.
+    1. If |init| [=map/contains=] none of "{{URLPatternInit/protocol}}", "{{URLPatternInit/hostname}}", "{{URLPatternInit/port}}", "{{URLPatternInit/pathname}}", "{{URLPatternInit/search}}", and "{{URLPatternInit/hash}}", then set |result|["{{URLPatternInit/hash}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/fragment=] and |type|.
+  1. If |init|["{{URLPatternInit/protocol}}"] [=map/exists=] then set |result|["{{URLPatternInit/protocol}}"] to the result of [=process protocol for init=] given |init|["{{URLPatternInit/protocol}}"] and |type|.
```
domenic, line 1794:
```suggestion
  1. If |init|["{{URLPatternInit/protocol}}"] [=map/exists=], then set |result|["{{URLPatternInit/protocol}}"] to the result of [=process protocol for init=] given |init|["{{URLPatternInit/protocol}}"] and |type|.
```

here and below feels slightly nicer to me.

domenic (approved):
LGTM with more nits, not all of which I'm sure about. @sisidovski did you want to take a look?

spec.bs
```diff
@@ -1685,23 +1770,33 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
-  1. If |init|["{{URLPatternInit/hostname}}"] is not null then set |result|["{{URLPatternInit/hostname}}"] to the result of [=process hostname for init=] given |init|["{{URLPatternInit/hostname}}"] and |type|.
-  1. If |init|["{{URLPatternInit/port}}"] is not null then set |result|["{{URLPatternInit/port}}"] to the result of [=process port for init=] given |init|["{{URLPatternInit/port}}"], |result|["{{URLPatternInit/protocol}}"], and |type|.
-  1. If |init|["{{URLPatternInit/pathname}}"] is not null:
+    1. If |init|["{{URLPatternInit/protocol}}"] does not [=map/exist=], then set |result|["{{URLPatternInit/protocol}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/scheme=] and |type|.
+    1. If |type| is not "`pattern`" and |init| [=map/contains=] none of "{{URLPatternInit/protocol}}", "{{URLPatternInit/hostname}}", "{{URLPatternInit/port}}" and "{{URLPatternInit/username}}", then set |result|["{{URLPatternInit/username}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/username=] and |type|.
```
sisidovski, line 1787: Just to confirm, |type| will be `pattern` only when constructing, right? match() and exec() also process URLPatternInit, but |type| is "url" in that case so I believe this is correct.

sisidovski (approved):

sisidovski:
Apologize for the delay, LGTM with a few questions.

spec.bs
jeremyroman: Added "when parsing a URL". You can actually supply a dictionary rather than a string if you want, and this behavior also applies there. e.g. `urlPattern.test({pathname: "/world"}, "https://user:pass@example.com/")` is a thing you can do.

spec.bs
```diff
@@ -510,17 +601,14 @@ A [=constructor string parser=] has an associated <dfn export for="constructor s
...
+  <p>In the constructor string algorithm, the pathname, search, and hash are wildcarded if earlier components are specified but later ones are not. For example, "`https://example.com/foo`" matches any search and any hash. Similarly, "`https://example.com`" matches any URL on that origin. This is analogous to the notion of a more specific component in the notes about [=process a URLPatternInit=] (e.g., a search is more specific than a pathname), but the constructor syntax only has a few cases where it is possible to specify a more specific component without also specifying the less specific components.
+  <p>The username and password components are always wildcard unless they are explicitly specified.
```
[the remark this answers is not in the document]
jeremyroman: Added a mention here.

spec.bs
jeremyroman: Yes, exactly.

jeremyroman:
Chromium CL has also landed, and should ship in M121. 🤞

Issue #191, opened 2023-10-06
Attribute to show the regular expression usage.


yoshisatoyanagisawa:
URLPattern are used in several proposals now:

- [Speculation rules](https://wicg.github.io/nav-speculation/speculation-rules.html#document-rule-url-pattern-predicate)
- [Service worker static routing](https://github.com/WICG/service-worker-static-routing-api#webidl)
- https://github.com/WICG/compression-dictionary-transport/issues/48

URLPattern supports regular expressions, but it is concerned to execute arbitrary user-provided regular expressions in trusted area in the browser due to security reasons.  It is suggested to prohibit regular expressions in some APIs using URLPattern:
- https://github.com/w3c/ServiceWorker/pull/1686/files
- https://github.com/WICG/compression-dictionary-transport/issues/42#issuecomment-1639774926

To avoid unexpected regular expression use in URLPattern, a new attribute to show the regular expression has been considered in https://github.com/WICG/urlpattern/issues/182#issuecomment-1716988239.  With this flag, web developers can understand unexpected regular expression usage by themselves.

Alternative solution for this is having an option to raise when the regular expression is used in the pattern because denying regular expression depends on APIs using URLPattern.  Some APIs may allow it and the other may not.  We should not have such a thing in the regular path.

Jamesernator:
The concern here is just ReDoS right? It seems unfortunate to limit patterns like `/(one|two)/`, would it be possible to instead of restricting *all* regexps to instead only limit those that contain potential backtracking?

domenic:
That's not the concern, actually. The issue is that several of these features run in browser processes (usually the network process) which don't have a JavaScript engine, and thus don't have a regular expression engine. Bringing regexp support to the network process would require bringing a whole copy of V8, which as you can imagine, is not trivial.

(I'll also note that browser security doesn't make a distinction of the sort you're discussing, between "dangerous" untrusted input and "non-dangerous" untrusted input. Any untrusted input at all falls afoul of [the rule of 2](https://chromium.googlesource.com/chromium/src/+/master/docs/security/rule-of-2.md), at least in Chromium.)

annevk:
This concern applies to server operators too, e.g., with compression dictionaries. This again leads me to think that having some kind of subset would be a good idea.

It seems like another way of stating the constraint here is that there's no interest in implementing a regular expression engine in a safe language.

Jamesernator:
> in implementing a regular expression engine in a safe language.

[...]

Issue #182, opened 2023-09-12
Add a "using URLPattern in other APIs" section


domenic:
We're seeing URLPattern being used in several proposals now:

- [Speculation rules](https://wicg.github.io/nav-speculation/speculation-rules.html#document-rule-url-pattern-predicate)
- [Service worker static routing](https://github.com/WICG/service-worker-static-routing-api#webidl)
- Hopefully, [compression dictionaries](https://github.com/WICG/compression-dictionary-transport/issues/48)
- Potentially in the future, its original motivation, [service worker scopes](https://github.com/WICG/urlpattern/blob/main/explainer.md#service-worker-api) or [web app manifest scopes](https://github.com/WICG/urlpattern/blob/main/explainer.md#web-app-manifest).

I think we need clearer guidance on how to design APIs that accept URL patterns, as we've started to see some slight divergence.

Scenarios to consider:

- JS APIs:
  - Should accept `URLPattern` objects
  - How can they accept raw inputs? Keep in mind that `URLPattern` has multiple constructor forms: `new URLPattern({ ...componentsDictionary, baseURL })`, `new URLPattern(patternString, baseURL)`, and then variants of each of those with another options argument.
  - My suggestion:
    - Such APIs should have a default base URL, and then accept either:
    - `{ ...componentsDictionary }` (which automatically gets the base URL),
    - `{ ...componentsDictionary, baseURL }` (overriding the base URL), or
    - `patternString` (which automatically gets the base URL).
  - Downsides of my suggestion:
    - You can't set the options (i.e. `ignoreCase`)
    - You can't combine a string with an overridden base URL
    - Depends on fixing #179 to give intuitive results for this kind of implicit base URL inheritance
- HTTP headers or other string-accepting APIs: should accept input strings, with some natural base URL automatically applied
- JSON APIs:
  - My suggestion (and what we implemented for speculation rules: the same as the "raw inputs" sub-question for JS APIs.

Other considerations: is there any place for pathname-only patterns? IMO probably not because anytime you match on pathname you should also allow matching on search, i.e. discriminating between URL patterns like `/products/*` and `/product?id=*` is not good. But it's come up a few times.

My proposal is to use the above rules for such situations, and enforce "pathname only" by checking that the resulting URL pattern's protocol/username/password/hostname/port are set to match the base URL's origin. Providing a spec helper for such a check would be a good idea, I think.

/cc @sisidovski @yoshisatoyanagisawa

pmeenan:
[...]

PR #199, opened 2023-11-21
Add a section on other specs integrating with URLPattern

jeremyroman:
A section which covers how other specifications should use URLPattern and how developer-facing APIs should work is added, along with helpful algorithms.

One of these is whether a pattern has regexp groups (which may require an ECMAScript regexp engine); a corresponding WebIDL attribute is added to expose this property to authors as well.

   * Google Chrome
   * [WebKit](https://github.com/whatwg/urlpattern/pull/199#issuecomment-1829418966)
   * https://chromium-review.googlesource.com/c/chromium/src/+/5054393
   * Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1504650, https://bugs.chromium.org/p/chromium/issues/detail?id=1504683
   * Deno: https://github.com/denoland/rust-urlpattern/issues/42
   * kenchris/urlpattern-polyfill: https://github.com/kenchris/urlpattern-polyfill/issues/125
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/content/issues/30640

***
<a href="https://whatpr.org/urlpattern/199.html" title="Last updated on Nov 30, 2023, 2:15 AM UTC (0dd8dc0)">Preview</a> | <a href="https://whatpr.org/urlpattern/199/ed205a4...0dd8dc0.html" title="Last updated on Nov 30, 2023, 2:15 AM UTC (0dd8dc0)">Diff</a>


jeremyroman:
This relates to #182 and #191.

spec.bs
```diff
@@ -438,6 +452,8 @@ A [=component=] has an associated <dfn for=component>regular expression</dfn>, a
 
 A [=component=] has an associated <dfn for=component>group name list</dfn>, a [=list=] of strings, which must be set upon creation.
 
+A [=component=] has an associated <dfn for=component>has regexp groups flag</dfn>, a [=boolean=], which must be set upon creation.
```
domenic, line 456: These days we're trying to move away from "flags" which can be off or on, toward just booleans which are true or false. Having something that's a boolean with flag in its name is a bit confusing. So I'd just say "has regexp groups".

spec.bs
```diff
@@ -1902,6 +1935,65 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   1. Return the result of running [=canonicalize a hash=] given |strippedValue|.
 </div>
 
+<h2 id=integrating>Integrating with other specs</h2>
```
domenic, line 1939: In streams we called the counterpart section "Using streams in other specifications", which I kind of like. "Use in other specifications" or "Using URL patterns in other specifications" seem nice.

spec.bs
```diff
@@ -1902,6 +1936,65 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+To accomplish this, specifications should accept {{URLPatternCompatible}} as an argument to an [=operation=] or [=dictionary member=], and process it using the following algorithm, using the appropriate [=environment settings object=]'s [=environment settings object/API base URL=] or equivalent.
+
+<div algorithm>
+  To <dfn for=URLPattern>build a {{URLPattern}} from a WebIDL value</dfn> {{URLPatternCompatible}} |input| given [=URL=] |baseURL|, perform the following steps:
+
+  1. If the [=specific type=] of |input| is {{URLPattern}}:
+    1. Return |input|.
+  1. Otherwise, if the [=specific type=] of |input| is {{URLPatternInit}}:
+    1. If |input|["`baseURL`"] does not [=map/exist=], set it to the [=URL serializer|serialization=] of |baseURL|.
+    1. Return the result of constructing a {{URLPattern/constructor(input, options)}} given |input|.
```
domenic, line 1964: This isn't super well-defined, and it'd be more well-founded if we factored out the constructor steps into separate algorithms and called them. (Maybe after doing a [new](https://webidl.spec.whatwg.org/#new) `URLPattern`.) This ties into a general issue where we prefer other specs not to operate directly on the public API, but that's a larger refactoring.

If we don't want to do that kind of work, then I think the best way to do this would be something like

> 1. Let _urlPattern_ be a [new](https://webidl.spec.whatwg.org/#new) `URLPattern`.
> 2. Perform the `new URLPattern(input, baseURL, options)` [constructor steps](https://webidl.spec.whatwg.org/#constructor-steps) with _urlPattern_ as [this](https://webidl.spec.whatwg.org/#this) and « _input_, the serialization of _baseURL_ » as argument values.

spec.bs
```diff
@@ -1902,6 +1936,65 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+
+To accomplish this, specifications should accept {{URLPatternCompatible}} as an argument to an [=operation=] or [=dictionary member=], and process it using the following algorithm, using the appropriate [=environment settings object=]'s [=environment settings object/API base URL=] or equivalent.
+
+<div algorithm>
+  To <dfn for=URLPattern>build a {{URLPattern}} from a WebIDL value</dfn> {{URLPatternCompatible}} |input| given [=URL=] |baseURL|, perform the following steps:
+
+  1. If the [=specific type=] of |input| is {{URLPattern}}:
+    1. Return |input|.
+  1. Otherwise, if the [=specific type=] of |input| is {{URLPatternInit}}:
+    1. If |input|["`baseURL`"] does not [=map/exist=], set it to the [=URL serializer|serialization=] of |baseURL|.
```
domenic, line 1963:
```suggestion
    1. If |input|["{{URLPatternInit/baseURL}}"] does not [=map/exist=], set it to the [=URL serializer|serialization=] of |baseURL|.
```

spec.bs
```diff
@@ -1902,6 +1936,65 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+    1. Return the result of constructing a {{URLPattern/constructor(input, options)}} given |input|.
+  1. Otherwise:
+    1. [=Assert=]: The [=specific type=] of |input| is {{USVString}}:
```
domenic, line 1966:
```suggestion
    1. [=Assert=]: The [=specific type=] of |input| is {{USVString}}.
```

spec.bs
```diff
@@ -1902,6 +1936,65 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+    1. Return the result of constructing a {{URLPattern/constructor(input, baseURL, options)}} given |input| and the [=URL serializer|serialization=] of |baseURL|.
+</div>
+
+This allows authors to concisely specify most patterns, and use the {{URLPattern/constructor|constructor}} to access uncommon options if necessary. The implicit use of the base URL is similar to, and consistent with, [[HTML]]'s [=parse a URL=] algorithm.
```
domenic, line 1970:
```suggestion
This allows authors to concisely specify most patterns, and use the <a lt="URLPattern/constructor">constructor</a> to access uncommon options if necessary. The implicit use of the base URL is similar to, and consistent with, [[HTML]]'s [=parse a URL=] algorithm.
```

to avoid the code font, maybe?

spec.bs
```diff
@@ -1902,6 +1936,65 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+<div algorithm>
+  To <dfn for=URLPattern>build a {{URLPattern}} from an Infra value</dfn> |rawPattern| given [=URL=] |baseURL|, perform the following steps.
+
+  1. Let |serializedBaseURL| be the [=URL serializer|serialization=] of |baseURL|.
+  1. If |rawPattern| is a [=string=], then:
+    1. Return the result of constructing a {{URLPattern}} using the {{URLPattern/URLPattern(input, baseURL)}} constructor steps given |rawPattern| and |serializedBaseURL|.
+  1. Otherwise, if |rawPattern| is a [=map=], then:
+    1. Let |init| be «[ "`baseURL`" → |serializedBaseURL| ]», representing a dictionary of type {{URLPatternInit}}.
+    1. [=map/For each=] |key| → |value| of |rawPattern|:
+      1. If |key| is not the [=identifier=] of a [=dictionary member=] of {{URLPatternInit}} or one of its [=inherited dictionaries=], |value| is not a [=string=], or the member's type is not declared to be {{USVString}}, then return null.
```
domenic: It's not 100% clear to me that every spec wants the kind of restrictive error handling we've used here. But I feel like we discussed this in the past when writing the speculation rules spec. I'm OK carrying it over for now, until someone raises an issue.

spec.bs
```diff
@@ -1902,6 +1936,65 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+JSON data formats which include URL patterns should mirror the behavior of <a href="#integrating-javascript">JavaScript APIs</a> and accept both:
+* an object which specifies the components required to construct a pattern
+* a string (in the constructor string syntax)
+
+If a specification has an [[INFRA]] value (e.g., after using [=parse a JSON string to an Infra value=]), use the following algorithm, using the appropriate base URL (by default, the URL of the JSON resource).
+
+<div algorithm>
+  To <dfn for=URLPattern>build a {{URLPattern}} from an Infra value</dfn> |rawPattern| given [=URL=] |baseURL|, perform the following steps.
+
+  1. Let |serializedBaseURL| be the [=URL serializer|serialization=] of |baseURL|.
```
domenic, line 1983: Needs a final return null, e.g. if _rawPattern_ is a number.

spec.bs
```diff
@@ -1902,6 +1936,65 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+    1. Let |init| be «[ "`baseURL`" → |serializedBaseURL| ]», representing a dictionary of type {{URLPatternInit}}.
+    1. [=map/For each=] |key| → |value| of |rawPattern|:
+      1. If |key| is not the [=identifier=] of a [=dictionary member=] of {{URLPatternInit}} or one of its [=inherited dictionaries=], |value| is not a [=string=], or the member's type is not declared to be {{USVString}}, then return null.
+
+        <div class="note">This will need to be updated if {{URLPatternInit}} gains members of other types.</div>
+      1. Set |init|[|key|] to |value|.
+    1. Return the result of constructing a {{URLPattern}} using the {{URLPattern/URLPattern(input, baseURL)}} constructor steps given |init|.
+</div>
+
+Specifications may wish to leave room in their formats to accept options for {{URLPatternOptions}}, override the base URL, or similar, since it is not possible to construct a {{URLPattern}} object directly in this case, unlike in a JavaScript API.
```
domenic, line 1996: An example might be nice, e.g. with `relative_to`.

spec.bs
```diff
@@ -1902,6 +1936,65 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+1. **Be clear about what URLs will be matched**. For instance, algorithms during fetching are likely to operate on URLs with no [=url/fragment=]. If so, the specification should be clear that this is the case, and may advise showing a developer warning if a pattern which cannot match (e.g., because it requires a non-empty fragment) is used.
+
+<h3 id=integrating-javascript>Integrating with JavaScript APIs</h3>
+
+JavaScript APIs should accept all of:
+* a {{URLPattern}} object
+* a dictionary-like object which specifies the components required to construct a pattern
+* a string (in the constructor string syntax)
+
+To accomplish this, specifications should accept {{URLPatternCompatible}} as an argument to an [=operation=] or [=dictionary member=], and process it using the following algorithm, using the appropriate [=environment settings object=]'s [=environment settings object/API base URL=] or equivalent.
```
domenic, line 1955: Maybe move this typedef into a snippet down here, since that's where it's applicable.

spec.bs
```diff
@@ -1902,6 +1936,65 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   1. Return the result of running [=canonicalize a hash=] given |strippedValue|.
 </div>
 
+<h2 id=integrating>Integrating with other specs</h2>
+
+To promote consistency on the web platform, other documents integrating with this specification should adhere to the following guidelines, unless there is good reason to diverge.
```
domenic, line 1941: These are really great.

spec.bs
```diff
@@ -1902,6 +1936,65 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+1. **Be clear about regexp groups**. Some APIs may benefit from only allowing URL patterns which do not [=URLPattern/has regexp groups|have regexp groups=], for example, because user agents are likely to implement them in a different thread or process from those executing author script, and because of security or performance concerns, a JavaScript engine would not ordinarily run there. If so, this should be clearly documented (with reference to {{URLPattern/hasRegExpGroups}}) and the operation should report an error as soon as possible (e.g., by throwing a JavaScript exception). If possible, this should be feature-detectable to allow for the possibility of this constraint being lifted in the future. Avoid creating different subsets of URL patterns without consulting the editors of this specification.
```
domenic, line 1945:
```suggestion
1. **Be clear about regexp groups**. Some APIs may benefit from only allowing URL patterns which do not [=URLPattern/has regexp groups|have regexp groups=], for example, because user agents are likely to implement them in a different thread or process from those executing author script, and because of security or performance concerns, a JavaScript engine would not ordinarily run there. If so, this should be clearly documented (with reference to [=URLPattern/has regexp groups=]) and the operation should report an error as soon as possible (e.g., by throwing a JavaScript exception). If possible, this should be feature-detectable to allow for the possibility of this constraint being lifted in the future. Avoid creating different subsets of URL patterns without consulting the editors of this specification.
```

spec.bs
domenic: Hmm, since [initialize](https://whatpr.org/urlpattern/199.html#urlpattern-initialize) already exists, it might be better to just call that directly.

spec.bs
jeremyroman: Done. A parallel change for the subheadings wasn't obvious, but I at least changed the anchors.

spec.bs
jeremyroman: Eh, sure. Once it's developer-facing it'll probably point at the developer-facing documentation which will probably be of the WebIDL attribute, but 🤷.

spec.bs
jeremyroman: Done.

spec.bs
jeremyroman: Done.

spec.bs
jeremyroman: Done.

spec.bs
jeremyroman: Done.

spec.bs
jeremyroman: threw a bunch of stuff at the wall while talking to you and managed to make it work, I think

spec.bs
jeremyroman: Done.

spec.bs
jeremyroman: Done.

spec.bs
```diff
@@ -1902,6 +1935,88 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+    1. Let |pattern| be a [=new=] {{URLPattern}} with |realm|.
+    1. Run [=initialize=] given |pattern|, |init|, null, and an empty [=map=].
+    1. Return |pattern|.
+  1. Otherwise:
+    1. [=Assert=]: The [=specific type=] of |input| is {{USVString}}.
+    1. Let |pattern| be a [=new=] {{URLPattern}} with |realm|.
+    1. Run [=initialize=] given |pattern|, |input|, the [=URL serializer|serialization=] of |baseURL|, and an empty [=map=].
+    1. Return |pattern|.
+
+  <div class="issue">Ideally we wouldn't need a realm here. If we extricate the URL pattern concept from the {{URLPattern}} interface, we won't anymore.</div>
```
domenic, line 1977:
```suggestion
  <p class="XXX">Ideally we wouldn't need a realm here. If we extricate the URL pattern concept from the {{URLPattern}} interface, we won't anymore.</p>
```

is what WHATWG uses

spec.bs
```diff
@@ -1902,6 +1935,88 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+
+  1. Let |serializedBaseURL| be the [=URL serializer|serialization=] of |baseURL|.
+  1. If |rawPattern| is a [=string=], then:
+    1. Let |pattern| be a [=new=] {{URLPattern}} with |realm|.
+    1. Run [=initialize=] given |pattern|, |rawPattern|, |serializedBaseURL|, and an empty [=map=].
+
+        <div class="note">It may become necessary in the future to plumb non-empty options here.</div>
+    1. Return |pattern|.
+  1. Otherwise, if |rawPattern| is a [=map=], then:
+    1. Let |init| be «[ "`baseURL`" → |serializedBaseURL| ]», representing a dictionary of type {{URLPatternInit}}.
```
domenic, line 2001:
```suggestion
    1. Let |init| be «[ "{{URLPatternInit/baseURL}}" → |serializedBaseURL| ]», representing a dictionary of type {{URLPatternInit}}.
```

spec.bs
```diff
@@ -1902,6 +1935,88 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+        <div class="note">A future version of this specification might also have a less strict mode, if that proves useful to other specifications.</div>
+      1. Set |init|[|key|] to |value|.
+    1. Let |pattern| be a [=new=] {{URLPattern}} with |realm|.
+    1. Run [=initialize=] given |pattern|, |init|, null, and an empty [=map=].
+
+        <div class="note">It may become necessary in the future to plumb non-empty options here.</div>
+    1. Return |pattern|.
+  1. Otherwise, return null.
+
+  <div class="issue">Ideally we wouldn't need a realm here. If we extricate the URL pattern concept from the {{URLPattern}} interface, we won't anymore.</div>
```
domenic, line 2015:
```suggestion
  <p class="XXX">Ideally we wouldn't need a realm here. If we extricate the URL pattern concept from the {{URLPattern}} interface, we won't anymore.</p>
```

spec.bs
```diff
@@ -1902,6 +1935,88 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+</div>
+
+Specifications may wish to leave room in their formats to accept options for {{URLPatternOptions}}, override the base URL, or similar, since it is not possible to construct a {{URLPattern}} object directly in this case, unlike in a JavaScript API. For example, [[SPECULATION-RULES]] accepts a "`relative_to`" key which uses the [=document base URL=] instead of the JSON resource's URL.
```
domenic, line 2018:
```suggestion
Specifications may wish to leave room in their formats to accept options for {{URLPatternOptions}}, override the base URL, or similar, since it is not possible to construct a {{URLPattern}} object directly in this case, unlike in a JavaScript API. For example, [[SPECULATION-RULES]] accepts a "`relative_to`" key which can be used to switch to using the [=document base URL=] instead of the JSON resource's URL.
```

spec.bs
```diff
@@ -1902,6 +1935,88 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+  1. Otherwise:
+    1. [=Assert=]: The [=specific type=] of |input| is {{USVString}}.
+    1. Let |pattern| be a [=new=] {{URLPattern}} with |realm|.
+    1. Run [=initialize=] given |pattern|, |input|, the [=URL serializer|serialization=] of |baseURL|, and an empty [=map=].
+    1. Return |pattern|.
+
+  <div class="issue">Ideally we wouldn't need a realm here. If we extricate the URL pattern concept from the {{URLPattern}} interface, we won't anymore.</div>
+</div>
+
+This allows authors to concisely specify most patterns, and use the <a constructor for="URLPattern">constructor</a> to access uncommon options if necessary. The implicit use of the base URL is similar to, and consistent with, [[HTML]]'s [=parse a URL=] algorithm.
```
domenic, line 1980: Optional nit here and below: WHATWG style generally uses `<cite>Specification name</cite>` inline and then `[[SPEC]]` at the end of the paragraph, after the period. (So e.g. `<cite>HTML</cite>`, `<cite>Infra</cite>` and `<cite>Speculation Rules</cite>`.) Example at the second note in https://url.spec.whatwg.org/#idna .

domenic (commented):
Found a few more nits, but it's looking quite nice to me.

It might be worth adding (either here or in a followup PR) a section on using URL patterns in HTTP fields, since that's another use case that's coming up in practice. Initial thoughts:

- Encourage just using strings for now, and not necessarily accepting dictionaries
- The algorithm can be a wrapper around "build a URLPattern from an Infra value" that is only ever called with a string
- Give them some help on figuring out exactly which base URL to use. Maybe consider both the request and response header cases?
- Maybe omit the realm argument and just use "an implementation-defined realm", because it's quite silly to make HTTP specs deal with realms? We can still have an XXX box, but I'd hate to have the realm as part of the public interface for this particular call site. (Related discussion regarding "implementation-defined realm": https://github.com/whatwg/infra/issues/625)

spec.bs
jeremyroman: Done.

spec.bs
jeremyroman: Done.

jeremyroman:
> Found a few more nits, but it's looking quite nice to me.
> 
> It might be worth adding (either here or in a followup PR) a section on using URL patterns in HTTP fields, since that's another use case that's coming up in practice. Initial thoughts:
> 
> * Encourage just using strings for now, and not necessarily accepting dictionaries
> * The algorithm can be a wrapper around "build a URLPattern from an Infra value" that is only ever called with a string
> * Give them some help on figuring out exactly which base URL to use. Maybe consider both the request and response header cases?
> * Maybe omit the realm argument and just use "an implementation-defined realm", because it's quite silly to make HTTP specs deal with realms? We can still have an XXX box, but I'd hate to have the realm as part of the public interface for this particular call site. (Related discussion regarding "implementation-defined realm": ["Parse JSON to Infra" algorithms shouldn't require a current JS realm infra#625](https://github.com/whatwg/infra/issues/625))

Yeah, I thought about including HTTP fields in this PR, but wasn't sure if we'd hammered out enough of the details yet.

At surface I would have thought the constructor string syntax would have been the most natural fit, but from what I understand of WICG/compression-dictionary-transport#52 they're currently looking at separate `match-path` and `match-search` (or `match-query`) params (contrary to your proposal here), and it's not clear to me if their reasons for doing so generalize. Like, here, it's somewhat pointless to match on things like protocol and port, but that's not necessarily the case for most hypothetical future header fields.

Besides that one I don't have a clear enough concept of the likely uses to be confident writing guidance, and would probably rather they file an issue or otherwise bring it up to us for now.

domenic:
> At surface I would have thought the constructor string syntax would have been the most natural fit, but from what I understand of [WICG/compression-dictionary-transport#52](https://github.com/WICG/compression-dictionary-transport/issues/52) they're currently looking at separate `match-path` and `match-search` (or `match-query`) params (contrary to your proposal here), and it's not clear to me if their reasons for doing so generalize. Like, here, it's somewhat pointless to match on things like protocol and port, but that's not necessarily the case for most hypothetical future header fields.

Hmm, we should discuss with them what's going on. This seems contrary to the design of URL patterns...

spec.bs
```diff
@@ -1902,6 +1935,88 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+    1. Run [=initialize=] given |pattern|, |init|, null, and an empty [=map=].
+
+        <div class="note">It may become necessary in the future to plumb non-empty options here.</div>
+    1. Return |pattern|.
+  1. Otherwise, return null.
+
+  <p class="XXX">Ideally we wouldn't need a realm here. If we extricate the URL pattern concept from the {{URLPattern}} interface, we won't anymore.</p>
+</div>
+
+Specifications may wish to leave room in their formats to accept options for {{URLPatternOptions}}, override the base URL, or similar, since it is not possible to construct a {{URLPattern}} object directly in this case, unlike in a JavaScript API. For example, Speculation Rules accepts a "`relative_to`" key which can be used to switch to using the [=document base URL=] instead of the JSON resource's URL. [[SPECULATION-RULES]]
```
domenic, line 2018: I'd say either `speculation rules` or `<cite>Speculation Rules</cite>`

domenic (approved):
LGTM. It'd be great to get @yoshisatoyanagisawa's review as well, as someone who would be working on the service worker static routing spec, which follows these guidelines.

spec.bs
jeremyroman: Err, meant to add `<cite>` and simply forgot to.

jeremyroman:
While I'm here, want to note -- the `ignoreCase` option has the potential to be an obstacle to the cases which can't handle regexp groups, but I think it's actually fine because I think the match algorithm ends up making everything ASCII. Otherwise, the fact that `vi` regexps do Unicode case folding (so k could match not only uppercase K but also kelvin symbol K , for instance) would be a potential issue for implementations that implemented matching without a regexp engine.

sisidovski (approved):
LGTM. I think the new sections explain well. SW static routing API would be able to follow this change.

jeremyroman:
For multi-implementer interest, @annevk is WebKit cool with this? You seemed supportive of at least #191 (which is the real behavior change here right now, since nothing depends on the #182 aspects as yet).

spec.bs
```diff
@@ -1902,6 +1935,88 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   1. Return the result of running [=canonicalize a hash=] given |strippedValue|.
 </div>
 
+<h2 id=other-specs>Using URL patterns in other specifications</h2>
+
+To promote consistency on the web platform, other documents integrating with this specification should adhere to the following guidelines, unless there is good reason to diverge.
+
+1. **Accept shorthands**. Most author patterns will be simple and straightforward. Accordingly, APIs should accept shorthands for those common cases and avoid the need for authors to take additional steps to transform these into complete {{URLPattern}} objects.
+1. **Respect the base URL**. Just as URLs are generally parsed relative to a base URL for their environment (most commonly, a [=document base URL=]), URL patterns should respect this as well. The {{URLPattern}} constructor itself is an exception because it directly exposes the concept itself, similar to how the <a interface spec=URL>URL</a> constructor does not respect the base URL even though the rest of the platform does.
```
yoshisatoyanagisawa, line 1943: "most commonly" might mean that each specification can choose the base URL to something most relevant to it, right?

spec.bs
```diff
@@ -1902,6 +1935,88 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+
+JavaScript APIs should accept all of:
+* a {{URLPattern}} object
+* a dictionary-like object which specifies the components required to construct a pattern
+* a string (in the constructor string syntax)
+
+To accomplish this, specifications should accept {{URLPatternCompatible}} as an argument to an [=operation=] or [=dictionary member=], and process it using the following algorithm, using the appropriate [=environment settings object=]'s [=environment settings object/API base URL=] or equivalent.
+
+<div algorithm>
+  To <dfn for=URLPattern>build a {{URLPattern}} from a WebIDL value</dfn> {{URLPatternCompatible}} |input| given [=/URL=] |baseURL| and [=ECMAScript/realm=]</a> |realm|, perform the following steps:
```
yoshisatoyanagisawa, line 1961: nit but where is a counterpart of this </a>?

spec.bs
```diff
@@ -1902,6 +1935,88 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+  1. Otherwise:
+    1. [=Assert=]: The [=specific type=] of |input| is {{USVString}}.
+    1. Let |pattern| be a [=new=] {{URLPattern}} with |realm|.
+    1. Run [=initialize=] given |pattern|, |input|, the [=URL serializer|serialization=] of |baseURL|, and an empty [=map=].
+    1. Return |pattern|.
+
+  <p class="XXX">Ideally we wouldn't need a realm here. If we extricate the URL pattern concept from the {{URLPattern}} interface, we won't anymore.</p>
+</div>
+
+This allows authors to concisely specify most patterns, and use the <a constructor for="URLPattern">constructor</a> to access uncommon options if necessary. The implicit use of the base URL is similar to, and consistent with, <cite>HTML</cite>'s [=parse a URL=] algorithm. [[HTML]]
```
yoshisatoyanagisawa, line 1980: nit but this phrase reminds me https://github.com/whatwg/urlpattern/issues/200.  It seems that the URLPattern initialization for the short-hand is slightly different from the URL algorithm.

yoshisatoyanagisawa:
Thanks for looping me in, @domenic.
LGTM.  I left some minor comments.

spec.bs
jeremyroman: Below it suggests the environment settings object's API base URL, which is the document base URL most of the time (in a document), but in a worker it's the global scope's base URL. In an HTTP header it would probably be the request or response URL. Most of the time there's a fairly strong precedent about what URLs are resolved relative to already, and specs should probably follow that unless they have a compelling reason to deviate.

spec.bs
jeremyroman: I don't understand what you're asking; can you elaborate?

spec.bs
jeremyroman: It is indeed slightly different, yeah. Still, I think it's good to keep it broadly similar where we can.

jeremyroman:
> Given the comments like
> 
> > Thanks. I'll have to wait for the PR to get resolved before it is something I could reference but as it looks right now, URLPatternCompatible is specific to the Javascript API.
> 
> over in [WICG/compression-dictionary-transport#52](https://github.com/WICG/compression-dictionary-transport/issues/52) it seems like it'd be good to add a section on HTTP header usage, either in this PR or as a near-future followup.

I agree it would be good to cover once we're agreed on what to do and feel fairly confident it will generalize. I was hoping to avoid entangling these two, so doing HTTP header fields in a followup. It feels slightly silly to write our opinions here and then turn around and use ourselves as authority to back up our own position, but if that's helpful we can do that too.

annevk:
@jeremyroman I don't understand the references to Deno in https://github.com/whatwg/urlpattern/pull/199#issuecomment-1826056297.

WebKit is supportive of figuring out an interface to URLPattern so it can be reused in environments where regular expression engines are a no-no.

I'm not entirely sure that the JavaScript API needs to make the difference observable (as that won't be accessible in those environments), but I guess it's reasonable.

jeremyroman:
> @jeremyroman I don't understand the references to Deno in [#199 (comment)](https://github.com/whatwg/urlpattern/pull/199#issuecomment-1826056297).

Those were supposed to be links to issues in this repository; for some reason Github autolinked to the wrong repository. (???)

> WebKit is supportive of figuring out an interface to URLPattern so it can be reused in environments where regular expression engines are a no-no.
> 
> I'm not entirely sure that the JavaScript API needs to make the difference observable (as that won't be accessible in those environments), but I guess it's reasonable.

One reason might be registering service worker static routes, where the pattern is registered from a JavaScript environment but evaluated in the fetch handling path.

annevk:
Makes sense, yeah. I do think we need a string-only version (realm-free) of the part you're not explicitly seeking feedback on yet, but we can get there through iteration.

spec.bs
yoshisatoyanagisawa: Oops, I should have updated this when I noticed the comment can only be given per line X(

In this line, there is `</a>`.  However, I could not find opening of this `</a>`.

spec.bs
```diff
@@ -450,7 +466,24 @@ A [=component=] has an associated <dfn for=component>group name list</dfn>, a [=
...
   1. Let |pattern string| be the result of running [=generate a pattern string=] given |part list| and |options|.
-  1. Return a new [=component=] whose [=component/pattern string=] is |pattern string|, [=component/regular expression=] is |regular expression|, and [=component/group name list=] is |name list|.
+  1. Let |has regexp groups| be false.
+  1. [=list/For each=] |part| of |part list|:
+    1. If |part|'s [=part/type=] is "<a for=part/type>`regexp`</a>", then set |has regexp groups| to true.
+  1. Return a new [=component=] whose [=component/pattern string=] is |pattern string|, [=component/regular expression=] is |regular expression|, [=component/group name list=] is |name list|, and [=component/has regexp groups=] is |has regexp groups|.
+</div>
+
+<div algorithm>
+  A {{URLPattern}} |pattern| <dfn for=URLPattern>has regexp groups</dfn> if the following steps return true:
```
domenic, line 476:
```suggestion
  A {{URLPattern}} |pattern| <dfn export for=URLPattern>has regexp groups</dfn> if the following steps return true:
```

spec.bs
domenic, line 1961:
```suggestion
  To <dfn export>build a {{URLPattern}} from a Web IDL value</dfn> {{URLPatternCompatible}} |input| given [=/URL=] |baseURL| and [=ECMAScript/realm=]</a> |realm|, perform the following steps:
```

spec.bs
```diff
@@ -1902,6 +1935,88 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+<h3 id=other-specs-json>Integrating with JSON data formats</h3>
+
+JSON data formats which include URL patterns should mirror the behavior of <a href="#other-specs-javascript">JavaScript APIs</a> and accept both:
+* an object which specifies the components required to construct a pattern
+* a string (in the constructor string syntax)
+
+If a specification has an Infra value (e.g., after using [=parse a JSON string to an Infra value=]), use the following algorithm, using the appropriate base URL (by default, the URL of the JSON resource). [[INFRA]]
+
+<div algorithm>
+  To <dfn for=URLPattern>build a {{URLPattern}} from an Infra value</dfn> |rawPattern| given [=/URL=] |baseURL| and [=ECMAScript/realm=] |realm|, perform the following steps.
```
domenic, line 1991:
```suggestion
  To <dfn export>build a {{URLPattern}} from an Infra value</dfn> |rawPattern| given [=/URL=] |baseURL| and [=ECMAScript/realm=] |realm|, perform the following steps.
```

domenic (commented):
You should also export the "match" definition https://urlpattern.spec.whatwg.org/#match

spec.bs
jeremyroman: Done.

spec.bs
jeremyroman: It was an mistakenly left behind when rewriting back and forth between Bikeshed shorthand and `<a>` syntax. Thanks for catching; removed.

spec.bs
jeremyroman: Done.

spec.bs
jeremyroman: Done.

jeremyroman:
@yoshisatoyanagisawa @domenic still lgty both? I'll merge if so, as I think we've met the process requirements.

domenic:
Still LGTM except I don't think you saw this comment:

> You should also export the "match" definition https://urlpattern.spec.whatwg.org/#match

spec.bs
yoshisatoyanagisawa: ok. thanks for the clarification.

spec.bs
yoshisatoyanagisawa: Ok.  I hope we can discuss about the topic in #200.

yoshisatoyanagisawa:
Thanks for heads up.
Still lgtm.

jeremyroman:
> Still LGTM except I don't think you saw this comment:
> 
> > You should also export the "match" definition https://urlpattern.spec.whatwg.org/#match

Whoops, super easy to miss unfortunately because of how GitHub requires me to switch to Files to be able to start a review (and not send tons of individual emails), which hides comments not attached to lines. Will fix.

PR #203, opened 2024-01-12
Correct processing of non-string members of a base URL record

### LINKED ISSUE #202: Incorrect handling of base URL in URLPatternInit processing

A correct implementation of the current spec fail with this kind of simple input:

```js
new URLPattern({ pathname: '/books/:id', baseURL: 'https://example.com' });
```

The issue comes in part from the use of URL spec's [internal parsers](https://url.spec.whatwg.org/#url-parsing). These parsers manipulate [URL records](https://url.spec.whatwg.org/#url-representation) with some fields (host, port, query, fragment) that default to `null` when empty.

When those empty fields are then used to fill the default url pattern (step 11-7 to 11-10 of [*process a URLPatternInit*](https://urlpattern.spec.whatwg.org/#process-a-urlpatterninit)), the [*processing a base URL string*](https://urlpattern.spec.whatwg.org/#process-a-base-url-string) assertion fail (*Assert: input is not null.*)

In the special case of the baseURL port, URL record use unsigned integers and not strings, so the assertion pass but the code fail further down in [escape a pattern string](https://urlpattern.spec.whatwg.org/#escape-a-pattern-string) (*1. Assert: input is an ASCII string.*).

As per Web IDL, implicit string conversion would yield `"null"` for those empty field (as the URL record struct is not marked `[LegacyNullToEmptyString]`) which is not what we want.

The spec should check if a baseURL field is not `null` before using it to fill the url pattern. As for the port it should be converted to a string but is the processing of the stringified port really necessary as we are certain its only ASCII characters and there's nothing to escape ?).

jeremyroman:
Some of these values may be null (or in the case of port, an integer) and
must be converted to strings before they can be applied to URLPatternInit.

The conversion of the port to an integer is also necessary in the encoding
callbacks.

Addresses #202, at least in part.

   * Google Chrome
   * n/a (change is simply fixing a bug in a non-controversial way)
   * n/a (this is a fix to how the spec expresses this, but does not represent a behavior change, and existing tests cover this behavior)
   * Chromium: n/a (believed to work correctly in Chromium)
   * Deno: n/a (no reason to believe a change is required)
   * kenchris/urlpattern-polyfill: n/a (no reason to believe a change is required)
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: n/a (change is on a spec detail not expressly documented)

***
<a href="https://whatpr.org/urlpattern/203.html" title="Last updated on Jan 16, 2024, 7:21 PM UTC (e399ab0)">Preview</a> | <a href="https://whatpr.org/urlpattern/203/531a2cc...e399ab0.html" title="Last updated on Jan 16, 2024, 7:21 PM UTC (e399ab0)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1740,7 +1740,7 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
      <p class="note">Note, we set the [=URL record=]'s [=url/scheme=] in order for the [=basic URL parser=] to recognize and normalize default port values.</p>
   1. Let |parseResult| be the result of running [=basic URL parser=] given |portValue| with |dummyURL| as <i>[=basic URL parser/url=]</i> and [=port state=] as <i>[=basic URL parser/state override=]</i>.
   1. If |parseResult| is failure, then throw a {{TypeError}}.
-  1. Return |dummyURL|'s [=url/port=] or empty string if it is null.
+  1. Return |dummyURL|'s [=url/port=], [=serialize an integer|serialized=], or empty string if it is null.
 </div>
 
 <div algorithm>
--- a/spec.bs
+++ b/spec.bs
@@ -18,6 +18,9 @@ spec: ECMASCRIPT; urlPrefix: https://tc39.es/ecma262/
   type: dfn
     text: IdentifierPart; url: #prod-IdentifierPart
     text: IdentifierStart; url: #prod-IdentifierStart
+spec: URL; urlPrefix: https://url.spec.whatwg.org/
+  type: dfn
+    text: serialize an integer; url: #serialize-an-integer
 </pre>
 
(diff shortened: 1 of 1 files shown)


sisidovski (approved):
Looks good to me, thanks for taking this!

jeremyroman:
@annevk I think we need a second implementer interest even for little fixes like this, right? Is WebKit cool with this?

annevk:
I think for cases where you are just fixing a bug in the standard when there are no conflicting implementations out there you can use your best judgment and carry on.

Though perhaps the extra waiting time helps as I see you fixed another another nit. :-)

Anyway, looks good.

jeremyroman:
Thanks Anne, noted for future changes in this vein.

PR #232, opened 2024-08-28
Editorial: Add brief explanations to examples

### LINKED ISSUE #229: Make it clearer that parentheses are needed to use regex's

It wasn't clear to me that parentheses are used to create regex groups.

It is kind of explained with examples like this:

> A regular expression can also be used instead, so the pathname pattern "/blog/:year(\\d+)/:month(\\d+)" will match "/blog/2012/02".

Or bits like this:

> "/(bar)"
>     "[char](https://urlpattern.spec.whatwg.org/#token-type-char)" and "[regexp](https://urlpattern.spec.whatwg.org/#token-type-regexp)" [tokens](https://urlpattern.spec.whatwg.org/#token).

But given parentheses are commonly used in regexs themselves I think this could be made clearer. Unless I missed where this was explained?

jeremyroman:
These examples form the only quick reference material in the spec. While MDN goes into more depth, brief explanations of what is happening in this syntax don't take up much room and may make it easier to tell what's going on, especially with regexp parts.

Resolves #229.

***
<a href="https://whatpr.org/urlpattern/232.html" title="Last updated on Aug 29, 2024, 3:11 PM UTC (54ab810)">Preview</a> | <a href="https://whatpr.org/urlpattern/232/d13ebea...54ab810.html" title="Last updated on Aug 29, 2024, 3:11 PM UTC (54ab810)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -76,6 +76,8 @@ It can be constructed using a string for each component, or from a shorthand str
     * `http://example.com/products/`
     * `https://example.com:8443/blog/our-greatest-product-ever`
   </ul>
+
+  This is a fairly simple pattern which requires most components to either match an exact string, or allows any string ("`*`"). The [=URL pattern/pathname component=] matches any path with at least two `/`-separated path components, the first of which is captured as "`category`".
 </div>
 
 <div class="example" id="example-intro-2">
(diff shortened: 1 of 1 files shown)


jeremyroman:
@sisidovski @tunetheweb wdyt?

sisidovski (approved):
Not sure this resolves #229, but the PR itself LGTM.

sisidovski:
IMO adding an explicit explanation paragraph how to create a custom regexp group in [2. Pattern strings](https://urlpattern.spec.whatwg.org/#pattern-strings) makes sense. WDYT?

jeremyroman:
> IMO adding an explicit explanation paragraph how to create a custom regexp group in [2. Pattern strings](https://urlpattern.spec.whatwg.org/#pattern-strings) makes sense. WDYT?

Added the text "enclosed in parentheses" to that section.

Issue #240, opened 2024-12-27
Why does this test succeed with pathname "/foo/bar"?


anonrig:
### What is the issue with the URL Pattern Standard?

Referencing a WPT test:

```
  {
    "pattern": [{ "pathname": "./foo/bar", "baseURL": "https://example.com" }],
    "inputs": [{ "pathname": "foo/bar", "baseURL": "https://example.com" }],
    "exactly_empty_components": [ "port" ],
    "expected_obj": {
      "pathname": "/foo/bar"
    },
    "expected_match": {
      "protocol": { "input": "https", "groups": {}},
      "hostname": { "input": "example.com", "groups": {}},
      "pathname": { "input": "/foo/bar", "groups": {}}
    }
  },
```

canonicalize_pathname method adds `/-` and removes `/-` from the method for several reasons. Ref: https://urlpattern.spec.whatwg.org/#canonicalize-a-pathname

Why does the expected_obj has a pathname of `/foo/bar` but not `./foo/bar`?

cc @annevk  @jeremyroman would you mind helping me understand the behavior here?

jeremyroman:
Because `./` path components are ignored in [URL path parsing](https://url.spec.whatwg.org/#path-state), so the pathname is simply the result of resolving `foo/bar` relative to the `/` implied in the base URL, which is `/foo/bar`.

```js
new URL('./foo/bar', 'https://example.com').toString()
// 'https://example.com/foo/bar'
```

anonrig:
@jeremyroman You're right but canonicalize pathname function prepends /- to the input and disables URL path parsing for this particular edge case.

Here's the ref for url pattern pathname canonicalize: https://urlpattern.spec.whatwg.org/#canonicalize-a-pathname

Particularly this hack should be removed:

> The URL parser will automatically prepend a leading slash to the canonicalized pathname. This does not work here unfortunately. This algorithm is called for pieces of the pathname, instead of the entire pathname, when used as an encoding callback. Therefore we disable the prepending of the slash by inserting our own. An additional character is also inserted here in order to avoid inadvertantly collapsing a leading dot due to the fake leading slash being interpreted as a "/." sequence. These inserted characters are then removed from the result below.

Using URL parser example, - character is not removed:

```
> new URL('/-./foo/bar', 'https://example.com').toString()
'https://example.com/-./foo/bar'
```

jeremyroman:
Right, and `URLPattern` matches that:

```
[...]

PR #243, opened 2025-01-06
Correct condition for opaque paths in base URL

anonrig:
This fixes one of the test conditions. Relevant Ada URL change https://github.com/ada-url/ada/pull/785/commits/da9a99118b63b60a67f41abf348daf0025783df0

Fixes https://github.com/whatwg/urlpattern/issues/240

cc @jeremyroman @sisidovski

### DIFF
--- a/review-drafts/2024-03.bs
+++ b/review-drafts/2024-03.bs
@@ -1868,7 +1868,7 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
     1. If the following are all true:
       <ul>
         <li>|baseURL| is not null;</li>
-        <li>|baseURL| has an [=url/opaque path=]; and</li>
+        <li>|baseURL| does not have an [=url/opaque path=]; and</li>
         <li>the result of running [=is an absolute pathname=] given |result|["{{URLPatternInit/pathname}}"] and |type| is false,
       </ul>
       <p>then:


jeremyroman:
Thanks very much. A couple minor procedural things before I merge this:

1. Please tweak the commit message according to [WHATWG style](https://whatwg.org/ipr-policy#31-contributors-and-workstream-participants). It's a little debatable, but I'd consider this editorial because all existing implementations and the tests already work in the way this is corrected to, and was apparently intended to, read. So, something like "Editorial: Correct condition for opaque paths in base URL", and add some text to the commit description clarifying that it was inverted, and this makes it match existing implemented and tested behavior.
2. It looks like the IPR bot is recognizing you as having signed as an individual, but I think the [policy](https://whatwg.org/ipr-policy#31-contributors-and-workstream-participants) might require you to be recognized as a Cloudflare employee unless given an exception by the steering group. It looks like [Cloudflare does participate as an organization](https://github.com/whatwg/participant-data/blob/82358193c7775c4724fb3f0f7fd60537378f2163/entities.json#L255-L280), so it might be appropriate for you to get added to the [cloudflare-whatwg](https://github.com/cloudflare-whatwg) organization to reflect that. On the other hand, if you think individual participation _is_ appropriate here, let me double-check with internal experts.

sisidovski (approved):
LGTM w/ Jeremy's comment.

I'm not fully confident how we should handle _opaque path_ here, but for now this change makes sense and it matches existing implementations.

domenic:
FWIW we usually treat spec bug fixes as non-editorial, even if the fix is to match the existing intent.

annevk:
We treat it as non-editorial for the purposes of writing the commit message, but if tests and existing implementations are already aligned there's no need to go through the extra trouble of the template.

anonrig:
> Thanks very much. A couple minor procedural things before I merge this:
> 
> 1. Please tweak the commit message according to [WHATWG style](https://github.com/whatwg/meta/blob/main/COMMITTING.md). It's a little debatable, but I'd consider this editorial because all existing implementations and the tests already work in the way this is corrected to, and was apparently intended to, read. So, something like "Editorial: Correct condition for opaque paths in base URL", and add some text to the commit description clarifying that it was inverted, and this makes it match existing implemented and tested behavior.

I've updated the commit message and description. Let me know if you need any further changes.

> 2. It looks like the IPR bot is recognizing you as having signed as an individual, but I think the [policy](https://whatwg.org/ipr-policy#31-contributors-and-workstream-participants) might require you to be recognized as a Cloudflare employee unless given an exception by the steering group. It looks like [Cloudflare does participate as an organization](https://github.com/whatwg/participant-data/blob/82358193c7775c4724fb3f0f7fd60537378f2163/entities.json#L255-L280), so it might be appropriate for you to get added to the [cloudflare-whatwg](https://github.com/cloudflare-whatwg) organization to reflect that. On the other hand, if you think individual participation _is_ appropriate here, let me double-check with internal experts.

I've reached out to @kentonv to join the organization.

Issue #242, opened 2025-01-06
Process a URLPatternInit doesn't call basic URL parser on base URL


shannonbooth:
### What is the issue with the URL Pattern Standard?

[process a URLPatternInit](https://urlpattern.spec.whatwg.org/#process-a-urlpatterninit) does:

> Set baseURL to the result of [parsing](https://url.spec.whatwg.org/#concept-url-parser) init["[baseURL](https://urlpattern.spec.whatwg.org/#dom-urlpatterninit-baseurl)"].

Which has blob handling, which does not seem required. This seems like it should be changed to invoke the basic URL parser instead: https://url.spec.whatwg.org/#concept-basic-url-parser

shannonbooth:
Another editorial problem I noticed is that "process a URLPatternInit" does:

> 1. Let baseHost be baseURL’s [host](https://url.spec.whatwg.org/#concept-url-host).
> 2. If baseHost is null, then set baseHost to the empty string.
> 3. Set result["[hostname](https://urlpattern.spec.whatwg.org/#dom-urlpatterninit-hostname)"] to the result of [processing a base URL string](https://urlpattern.spec.whatwg.org/#process-a-base-url-string) given baseHost and type.

But URL's host is a concept, and is not a string. e.g it may be an ipv4 address which is a 32 bit integer. I believe the spec should be calling this on host's serialization instead. But I can split this into a different issue if preferred (I will probably run into a few of these, I am implementing the spec from scratch).

shannonbooth:
One more I noticed I can also split out into a different issue, but putting it down here for reference.

Parse a pattern string (https://urlpattern.spec.whatwg.org/#parse-a-pattern-string), step 3.9.1 does:

> 1. Set prefix be the result of running [consume text](https://urlpattern.spec.whatwg.org/#consume-text) given parser.

But not prefix is in scope, so I believe should be 'let'.

shannonbooth:
Noticed another thing (I will do a pass over this afterwards and clean up this issue).

Canonicalizing an opaque pathname https://urlpattern.spec.whatwg.org/#canonicalize-an-opaque-pathname

On step 3. does:
> 3. Set dummyURL’s [path](https://url.spec.whatwg.org/#concept-url-path) to the empty string.

But in this case the URL path is a list. So I guess this must means to set it to a list with one item of the empty string?

jeremyroman:
[...]

PR #248, opened 2025-01-08
Use the basic URL parser when parsing URLs

jeremyroman:
Blob handling is not required, since the resulting URL is not stored in any way that would make the blob handling visible. This is consistent with what the URL constructor and similar uses do.

This change should not be observable.

Partially addresses #242.

***
<a href="https://whatpr.org/urlpattern/248.html" title="Last updated on Jan 8, 2025, 10:03 PM UTC (8c660ed)">Preview</a> | <a href="https://whatpr.org/urlpattern/248/807258f...8c660ed.html" title="Last updated on Jan 8, 2025, 10:03 PM UTC (8c660ed)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -513,10 +513,10 @@ A <dfn>component</dfn> is a [=struct=] with the following [=struct/items=]:
     1. If |input| is a {{USVString}}:
       1. Let |baseURL| be null.
       1. If |baseURLString| was given, then:
-        1. Set |baseURL| to the result of [=basic URL parser|parsing=] |baseURLString|.
+        1. Set |baseURL| to the result of running the [=basic URL parser=] on |baseURLString|.
         1. If |baseURL| is failure, return null.
         1. [=list/Append=] |baseURLString| to |inputs|.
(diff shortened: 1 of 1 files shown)


jeremyroman:
This is not quite editorial but close to it as I think it's more or less just clarification and simplification.

jeremyroman:
@shannonbooth I think this should address your first comment in #242, right?

@sisidovski Please review.

shannonbooth:
Yup, this fixes that portion of the issue, thanks!

Though I will mention that the HTML spec uses the language: "basic URL parse" in text to differentiate the two types of parsing so that the reader can know which one is called without needing to navigate to the spec link (the difference in language here - lack of "basic") is what made me notice the issue). So I personally think it would be nice to specifically say "basic URL parse" or similar, which would also be consistent with the canonicalization steps (e.g  https://urlpattern.spec.whatwg.org/#canonicalize-a-protocol)

jeremyroman:
Fair, I'll use "running the basic URL parser on _url_ [with _base_]" wording as used elsewhere here and in the URL living standard.

PR #250, opened 2025-01-08
Serialize base URL's host when a string is required

### LINKED ISSUE #242: Process a URLPatternInit doesn't call basic URL parser on base URL

jeremyroman:
If the URL's host is an IP address, the string representation is required.

This corrects/clarifies existing behavior and partially resolves #242.

***
<a href="https://whatpr.org/urlpattern/250.html" title="Last updated on Jan 8, 2025, 9:44 PM UTC (a5b70be)">Preview</a> | <a href="https://whatpr.org/urlpattern/250/807258f...a5b70be.html" title="Last updated on Jan 8, 2025, 9:44 PM UTC (a5b70be)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1858,7 +1858,7 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
     1. If |type| is not "`pattern`" and |init| [=map/contains=] none of "{{URLPatternInit/protocol}}", "{{URLPatternInit/hostname}}", "{{URLPatternInit/port}}" and "{{URLPatternInit/username}}", then set |result|["{{URLPatternInit/username}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/username=] and |type|.
     1. If |type| is not "`pattern`" and |init| [=map/contains=] none of "{{URLPatternInit/protocol}}", "{{URLPatternInit/hostname}}", "{{URLPatternInit/port}}", "{{URLPatternInit/username}}" and "{{URLPatternInit/password}}", then set |result|["{{URLPatternInit/password}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/password=] and |type|.
     1. If |init| [=map/contains=] neither "{{URLPatternInit/protocol}}" nor "{{URLPatternInit/hostname}}", then:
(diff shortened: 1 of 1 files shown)


sisidovski (approved):

spec.bs
```diff
@@ -1858,7 +1858,7 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
     1. If |type| is not "`pattern`" and |init| [=map/contains=] none of "{{URLPatternInit/protocol}}", "{{URLPatternInit/hostname}}", "{{URLPatternInit/port}}" and "{{URLPatternInit/username}}", then set |result|["{{URLPatternInit/username}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/username=] and |type|.
     1. If |type| is not "`pattern`" and |init| [=map/contains=] none of "{{URLPatternInit/protocol}}", "{{URLPatternInit/hostname}}", "{{URLPatternInit/port}}", "{{URLPatternInit/username}}" and "{{URLPatternInit/password}}", then set |result|["{{URLPatternInit/password}}"] to the result of [=processing a base URL string=] given |baseURL|'s [=url/password=] and |type|.
     1. If |init| [=map/contains=] neither "{{URLPatternInit/protocol}}" nor "{{URLPatternInit/hostname}}", then:
-      1. Let |baseHost| be |baseURL|'s [=url/host=].
+      1. Let |baseHost| be the [=host serializer|serialization=] of |baseURL|'s [=url/host=].
```
shannonbooth, line 1861: Sorry, I didn't happen to catch this one - but I don't believe the fix here is fully correct. URL's host may be null, and it is not defined in the URL spec how to serialize a null host. This also makes the second step not quite correct, as the serialization of host always returns an ASCII string, so can never be null.

Issue #220, opened 2024-02-21
Opaque Host and domain encoding


rubycon:
### What is the issue with the URL Pattern Standard?

Here's a test case from the WPT test suite, it's expected to succeed (Chrome 123 pass).

```js
const up1 = new URLPattern({ hostname : "xn--caf-dma.com" });
up1.exec({ hostname: "café.com" });
```

The test expect the hostname "café.com" to be converted to "xn--caf-dma.com" and to match the pattern.

But as per the spec, without a special scheme the host is [considered an opaque host](https://url.spec.whatwg.org/#concept-url-host) and not a domain name, therefore it should be percent-encoded to "caf%C3%A9.com" and not IDNA encoded to "xn--caf-dma.com". The pattern won't match.

Is it an oversight in the test suite/Chrome implementation or should hostname always be considered domain name ?

In some case the specs already allows opaque URL to be treated as special (cf. [process pathname for init](https://urlpattern.spec.whatwg.org/#process-pathname-for-init)) and default the URL processing to the *most common* (most expected??) behaviour. 

Should it be the case here too ? Or maybe only in some cases... Like forcing the URL to a special scheme when the host is not an address and [domain labels](https://url.spec.whatwg.org/#domain-label) are detected ?

rubycon:
After some more digging, most of the fail tests in this issue https://github.com/whatwg/urlpattern/issues/206 are related to this *opaque host* issue. I think the idea was that these test should (at least) fail when checking [forbidden domain code point](https://url.spec.whatwg.org/#forbidden-domain-code-point) in the host parser. This checking is only done in the URL parser for non opaque host.

As for the domain encoding issue, all these cases produce URL encoded hostname:

```js
new URLPattern({ hostname : "café.com" }); // hostname pattern string: "caf%C3%A9.com"
new URLPattern({ protocol: "https", hostname : "café.com" }); // hostname pattern string: "caf%C3%A9.com"
new URLPattern("https://café.com" }); // hostname pattern string: "caf%C3%A9.com"
```

The first one seems logic (except if we consider that every non-adress hostname with unknown protocol should be considered domain) but the other two should be IDNA encoded domain name since they are [non opaque/special URLs](https://url.spec.whatwg.org/#is-special). It seems everything that goes into the pattern compiler is parsed separately, without context, so the hostname is always considered opaque. We can see the right behaviour when the domain is passed in the base URL : 

```js
[...]

Issue #252, opened 2025-01-15
URLPattern use of URL record is ambiguous


youennf:
### What is the issue with the URL Pattern Standard?

URLPattern is using URL records, and passing some URL records to URL algorithms. This might lead to ambiguous behaviour.

For instance https://urlpattern.spec.whatwg.org/#canonicalize-a-hostname is passing the URL record to parse hostname (via https://url.spec.whatwg.org/#hostname-state).
It is unclear whether an URL record is a special URL or not, so it is unclear whether parsing of the hostname (https://url.spec.whatwg.org/#concept-host-parser) will lead to https://url.spec.whatwg.org/#concept-opaque-host-parser.

Looking at WPT tests and Chrome results, the expected behaviour is to consider that the url is special and hostname is not considered opaque.

The same pattern is done in other places of the spec (canonicalisation algorithms of port, pathname, search and hash for instance), though it is not clear to me whether this has side effects.

Discussing with @annevk, it is unexpected to pass URL records directly like done in https://urlpattern.spec.whatwg.org/#canonicalize-a-hostname. A future url spec refactoring might make that clear.

Ideally, the url given to the basic URL Parser should be an URL built by a URL parser.
Maybe that is what the spec should do.

sisidovski:
Thank you for filing this. I agree, the current spec doesn't specify the given URL is special or not and this should be explicit. Or even the current spec may be wrong because the default value of [scheme](https://url.spec.whatwg.org/#concept-url-scheme) is empty string, which is not a [special scheme](https://url.spec.whatwg.org/#special-scheme) so the URL wouldn't be [special](https://url.spec.whatwg.org/#is-special).

> Looking at WPT tests and Chrome results, the expected behaviour is to consider that the url is special and hostname is not considered opaque.

Yes.

> Ideally, the url given to the basic URL Parser should be an URL built by a URL parser.
Maybe that is what the spec should do.

Are you referring [URL parser](https://url.spec.whatwg.org/#concept-url-parser)? I see there is a comment in [URL parser](https://url.spec.whatwg.org/#concept-url-parser), saying _"Non-web-browser implementations only need to implement the basic URL parser"_. I assume some non-web-browser implementers don't implement this algorithm. Do you have any concern about just using the result of [basic URL parser](https://url.spec.whatwg.org/#concept-basic-url-parser), given "http://dummy.test" as an input and pass it as _dummyURL_ in https://urlpattern.spec.whatwg.org/#canonicalize-a-hostname?

[...]

PR #255, opened 2025-01-30
Add proper canonicalization of domain names

rubycon:
The current specs only handle host canonicalization for IP addresses and opaque hostnames but if an URL has a special scheme its host should be normalized as a domain name (IDNA processing).

Cf. https://github.com/whatwg/urlpattern/issues/220

   * At least 2 tests already exist in [WPT](https://github.com/web-platform-tests/wpt/blob/master/urlpattern/resources/urlpatterntestdata.json): Pattern: `[{ "hostname": "xn--caf-dma.com" }]`, inputs= `[{ "hostname": "café.com" }]`; Pattern: `["http://🚲.com/"]`, Expected: `{ "protocol": "http", "hostname": "xn--h78h.com", "pathname": "/" }`.

***
<a href="https://whatpr.org/urlpattern/255.html" title="Last updated on Mar 24, 2025, 4:00 PM UTC (38e80d1)">Preview</a> | <a href="https://whatpr.org/urlpattern/255/e84799f...38e80d1.html" title="Last updated on Mar 24, 2025, 4:00 PM UTC (38e80d1)">Diff</a>

### DIFF
--- a/review-drafts/2024-03.bs
+++ b/review-drafts/2024-03.bs
@@ -1868,7 +1868,7 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
     1. If the following are all true:
       <ul>
         <li>|baseURL| is not null;</li>
-        <li>|baseURL| has an [=url/opaque path=]; and</li>
+        <li>|baseURL| does not have an [=url/opaque path=]; and</li>
         <li>the result of running [=is an absolute pathname=] given |result|["{{URLPatternInit/pathname}}"] and |type| is false,
       </ul>
       <p>then:
--- a/review-drafts/2024-03.bs
+++ b/review-drafts/2024-03.bs
@@ -1868,7 +1868,7 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
     1. If the following are all true:
       <ul>
         <li>|baseURL| is not null;</li>
-        <li>|baseURL| does not have an [=url/opaque path=]; and</li>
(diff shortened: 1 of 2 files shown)


annevk:
I think this will be fixed by https://github.com/whatwg/urlpattern/issues/252. The dummy URL will have a special scheme.

rubycon:
Fully resolved by https://github.com/whatwg/urlpattern/pull/269

annevk:
Yeah, I think that would be reasonable.

sisidovski:
Agree, the current spec and tests don't consider opaque hosts, but ideally we should support it. 

>  non-empty non-special scheme as having opaque hosts

It sounds a good direction!

rubycon:
@annevk @sisidovski I've been thinking about this. Is their any use cases ? I mean is their any URL schemes host outside the special schemes, which will need to be processed as opaque host ?

Most of URL-like URIs (`mailto`, `irc`... and pretty much every custom scheme) expected their host to be IDNA processed because they rely on DNS. So let's say we want to create an URLPattern for a mail server with [non-ASCII char hostname](https://mailoji.com), it's still better to process the hostname as non-opaque for better matching. Furthermore they need to be explicited init from a struct

Do you know of cases were opaque host will be more suitable ? The advantage over non-opaque is non-desctructive (percent) encoding of the hostname, maybe that's required in some cases ?

annevk:
Well `new URL()` will use the opaque host parser for `mailto` and such. So consistency with that would be good I think.

Issue #260, opened 2025-03-07
port pattern canonicalization seems wrong after all


annevk:
### What is the issue with the URL Pattern Standard?

I'd like to revisit #239. cc @anonrig 

> `new URLPattern({ "protocol": "http", "port": "80 " })`

The expectation of this test is that the port gets elided because 80 is the default port for "`http`".

This is the set of steps:

1. https://urlpattern.spec.whatwg.org/#dom-urlpattern-urlpattern-input-options The constructor invokes initialize.
2. https://urlpattern.spec.whatwg.org/#urlpattern-initialize Initialize This invokes create.
3. Create invokes https://urlpattern.spec.whatwg.org/#process-a-urlpatterninit (notably with "pattern" as second argument)
4. In step 16 this invokes https://urlpattern.spec.whatwg.org/#process-port-for-init but note that it returns the value immediately due to _type_ being "pattern". So result["port"] is "80 " including trailing space.
5. Then we go back up and hit step 6 of https://urlpattern.spec.whatwg.org/#url-pattern-create but this won't elide the port either as it has trailing whitespace. (A naïve implementation might strip the whitespace here, which is maybe what happened when the test got written?)
6. Then we hit step 13 but at this point we no longer have protocol as context so port elision can no longer occur.

cc @jeremyroman cc @sisidovski

anonrig:
You're right. The spec should mention trim c0 whitespace before checking if protocol's port is equal to port. Here's the Ada implementation that follows the test but fails to follow the specification:

```
  // If processedInit["protocol"] is a special scheme and processedInit["port"]
  // is a string which represents its corresponding default port in radix-10
  // using ASCII digits then set processedInit["port"] to the empty string.
  // TODO: Optimization opportunity.
  if (scheme::is_special(*processed_init->protocol)) {
    std::string_view port = processed_init->port.value();
    helpers::trim_c0_whitespace(port);
    if (std::to_string(scheme::get_special_port(*processed_init->protocol)) ==
        port) {
      processed_init->port->clear();
    }
  }
```

I recommend adding trim whitespace behavior to the spec itself.

sisidovski:
Yeah I believe the current wpt is the intuitive behavior for developers.

Maybe we can add the step removing c0 whitespace from the input in https://urlpattern.spec.whatwg.org/#process-port-for-init.

annevk:
[...]

PR #261, opened 2025-03-12
Remove leading and trailing spaces in the process port for init

sisidovski:
This will address https://github.com/whatwg/urlpattern/issues/260.

***
<a href="https://whatpr.org/urlpattern/261.html" title="Last updated on Mar 12, 2025, 11:35 AM UTC (1dc7e72)">Preview</a> | <a href="https://whatpr.org/urlpattern/261/c289ec2...1dc7e72.html" title="Last updated on Mar 12, 2025, 11:35 AM UTC (1dc7e72)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1944,6 +1944,7 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
 <div algorithm>
   To <dfn>process port for init</dfn> given a string |portValue|, a string |protocolValue|, and a string |type|:
 
+  1. Remove any leading and trailing [=C0 control or space=] from |portValue|.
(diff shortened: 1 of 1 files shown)


anonrig (commented):
LGTM

annevk:
This does mean that the "issue" can still be reproduced with "`8\t0`" as input, right? I wonder if we should perhaps do the opposite here and not trim and just change the test expectation.

sisidovski:
Yeah that's a fair point. We can solve "80 " easily but yes this only solves leading/trailing cases. Initially I thought this is better than nothing, but only fixing this case will introduce further confusions, changing the test expectation makes sense.

sisidovski:
OK, I'll work on fixing test expectation side. Let me close this PR.

Issue #266, opened 2025-03-20
port canonicalization does not return failure for port of 'invalid80'


shannonbooth:
### What is the issue with the URL Pattern Standard?

From what I understand, the following WPT test: https://github.com/web-platform-tests/wpt/blob/11d0a493dc7f9552ce1f99da80bb6e9238a60510/urlpattern/resources/urlpatterntestdata.json#L1235

Fails on a spec conforming implementation. This is because the basic URL parser with state override as 'port state' does not fail in: ["canonicalize a port"](https://urlpattern.spec.whatwg.org/#canonicalize-a-port).

It instead sets the port to "null" and returns due to the following sequence:

 * Step 1 of port state: code point of 'I' is not an ASCII digit
 *  Step 2 of port state: state override is provided
 * Step 2.1 of port state: buffer is the empty string
 * Step 2.2 of port state: state override is given, so URL parser returns.

As an idea of a potential fix, step 2.2 of port state parser in URL spec could be changed to return failure if code point is not EOF.

cc @annevk

annevk:
That sounds reasonable.

And sigh, we prolly need to verify all the state override cases.

annevk:
Actually, the use of opaque path state as an override is weird as well.

annevk:
Not just weird, it's broken as that's not a supported state override.

annevk:
@shannonbooth I looked at patching the URL standard for the port case specifically, but it does raise the question what we want to do for inputs such as

- `80/x`
- `80?x`
- `80†`

Should those be errors or not? Depends? I guess I don't mind treating them all as errors if everyone is okay with that, but we should add more tests.

cc @anonrig @sisidovski

sisidovski:
Thank you for filing this. Sounds reasonable to me.

> Which would be more consistent with how the hostname override works so let me prepare a patch against the URL standard that aligns with this.

Are proposed test cases consistent with the result of the URL port setter, right? If so I think we should have them in our test cases.

sisidovski:
> As an idea of a potential fix, step 2.2 of port state parser in URL spec could be changed to return failure if code point is not EOF.

So the proposed plan is to make the basic url parser returns failure. So we also should update the test to expect  "error" right?

annevk:
Yeah, see https://github.com/web-platform-tests/wpt/pull/51505 for the tests and https://github.com/whatwg/url/pull/866 for the URL standard change to address the port override state. Review welcome.

[...]

PR #268, opened 2025-03-20
Override states path start, query, and fragment do not return failure

annevk:
This came up in #266.


sisidovski (approved):
LGTM, thank you!

PR #269, opened 2025-03-24
Make dummyURL from the basic URL parser with a special scheme

sisidovski:
Based on the discussion in https://github.com/whatwg/urlpattern/issues/252.

We should not use URL records directly to canonicalize components, we should specify a special scheme as a default scheme. When we create a dummy URL, it should be created from the basic URL parser. This is not ideal but this can be achieved by creating a dummy URL with this fixed URL string.

***
<a href="https://whatpr.org/urlpattern/269.html" title="Last updated on Mar 25, 2025, 2:08 PM UTC (9105e2f)">Preview</a> | <a href="https://whatpr.org/urlpattern/269/5c979a3...9105e2f.html" title="Last updated on Mar 25, 2025, 2:08 PM UTC (9105e2f)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1721,7 +1721,7 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   To <dfn>canonicalize a username</dfn> given a string |value|:
 
   1. If |value| is the empty string, return |value|.
-  1. Let |dummyURL| be the result of [=create a dummy URL=].
+  1. Let |dummyURL| be the result of [=creating a dummy URL=].
   1. [=Set the username=] given |dummyURL| and |value|.
   1. Return |dummyURL|'s [=url/username=].
 </div>
@@ -1730,7 +1730,7 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   To <dfn>canonicalize a password</dfn> given a string |value|:
 
   1. If |value| is the empty string, return |value|.
-  1. Let |dummyURL| be the result of [=create a dummy URL=].
+  1. Let |dummyURL| be the result of [=creating a dummy URL=].
   1. [=Set the password=] given |dummyURL| and |value|.
   1. Return |dummyURL|'s [=url/password=].
 </div>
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1704,7 +1704,8 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
   To <dfn>canonicalize a protocol</dfn> given a string |value|:
 
   1. If |value| is the empty string, return |value|.
-  1. Let |dummyURL| be a new [=URL record=].
+  1. Let |dummyInput| be "http://dummy.test".
```
annevk, line 1707: It needs to be
```
"`http://...`"
```
so the string value ends up wrapped in a `code` element.

annevk: I think we should use `https://dummy.invalid/` as URL to more clearly indicate it's not meant to go anywhere and `https` because that's preferred over `http`.

Also, we should introduce "create a dummy URL" in the High-level operations section and invoke that here and elsewhere so the logic is more centralized.

spec.bs
```diff
@@ -1704,7 +1704,8 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
+  1. Let |dummyURL| be the result of running the [=basic URL parser=] on |dummyInput|.
   1. Let |parseResult| be the result of running the [=basic URL parser=] given |value| followed by "`://dummy.test`", with |dummyURL| as <i>[=basic URL parser/url=]</i>.
```
annevk, line 1709: Since we're not using state override I think we don't have to pass dummyURL either. This seems like an incorrect invocation of the basic URL parser at the moment.

annevk (commented):
Modulo a couple of nits I think this is the right approach, yeah. This is also pretty much how WebKit handles this.

spec.bs
sisidovski: Done. I'm a bit unsure where we have the create a dummy URL algorithm. I put it in the High-level operations section, but it might be better to put it in https://urlpattern.spec.whatwg.org/#urlpattern-internals?

sisidovski:
Updated based on your suggestions, thank you!

spec.bs
```diff
@@ -1704,18 +1711,17 @@ To <dfn>convert a modifier to a string</dfn> given a [=part/modifier=] |modifier
...
-  1. Return |dummyURL|'s [=url/scheme=].
+  1. Return |parseResult|'s [=url/scheme=].
 </div>
 
 <div algorithm>
   To <dfn>canonicalize a username</dfn> given a string |value|:
 
   1. If |value| is the empty string, return |value|.
-  1. Let |dummyURL| be a new [=URL record=].
+  1. Let |dummyURL| be the result of [=create a dummy URL=].
```
annevk, line 1724: You want "creating" not "create" here (and below) to get a proper sentence (you don't have to change the `<dfn>` for that, as I mentioned Bikeshed takes care of making that work).

sisidovski: My bad, fixed.

annevk:
@sisidovski when merging this, please take care to follow https://github.com/whatwg/meta/blob/main/COMMITTING.md (in particular the Squash and merge section). I noticed this went wrong a few times for this repository. Feel free to reach out on https://whatwg.org/chat if you have questions or ask @domenic internally.

